PR #38, opened 2014-03-31
multichar line sep on server socket

simula67:
This test verifies that the sockets can operate on
multichar line seperators for calls to get.

This should already have been verified by test 5.
Somehow that test passes when this one does not.
Test 5 seems to the on the client socket.

jnthn mentioned that MOAR does not like multichar
line sepearators. But the problem seems to persist
for Parrot and JVM backend in a slightly different
manner.

I only tested this patch on Linux.

### DIFF
--- a/S32-io/IO-Socket-INET.pl
+++ b/S32-io/IO-Socket-INET.pl
@@ -254,6 +254,31 @@ given $test {
             $sock.close();
         }
     }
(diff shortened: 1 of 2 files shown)


JJ:
Conflicts not solved, probably best to close it.

AlexDaniel:
Let us know if there's still an issue, preferably by filing a bug report.

PR #55, opened 2015-03-24
Rewrite parts of skipped tests for sequence with user class

usev6:


usev6:
I created a PR for this, so it can be discussed.

IMHO the old test code had some bugs in itself. I tried to rewrite without changing the intention of the tests.

As far as I understand it, there are at least two underlying problems with rakudo which surface here. (1) A defined multi candidate for infix:<cmp> is not known in sub SEQUENCE (/src/core/operators.pm) and therefore the checks for choosing .succ or .pred and -- probably -- matching of the endpoint of the sequence don't work as expected. (2) S03 say ```Likewise, if the given value or values are not numeric, .succ is assumed```. This doesn't seem to work that way.

PR #59, opened 2015-05-03
added RT ticket #s to skip/todo tickets

dnmfarrell:
I created ~600 RT tickets for rakudo skip/todo that were missing RT tickets. I've added the RT ticket #s back into the test suite here.

#nyc_hackathon!


coke:
@dnmfarrell - thanks for working on this. We have a few things to work on before we can apply this.

-#?rakudo todo "doesn't work yet due to copying of arrays"
+#?rakudo todo "doesn't work yet due to copying of arrays" RT #124650

But the RT and number needs to be inside the double quoted string, e.g.:

#?rakudo todo "doesn't work yet due to copying of arrays - RT #124650"

Additionally, the patch no longer cleanly applies. If you can resolve both these issues, I can get this applied.

Additionally, we need to close out some of these new RTs - we're not tracking issues about anything parrot right now (in fact, I removed these fudges last night), and for issues like 

124490
124491
124492

All these say "object type is NYI" - that's one ticket, not 3.

Until we get this patch applied, I'll start keeping a list of any merges/closes I make to your newly created tickets so we can keep things in sync.

moritz:
Also, please don't include the `#` character in fudge messages, since it appears in the TAP output, and it has special meaning there.

coke:
Moritz -we already do that - if it's not being escaped properly, that's on our test harness.

coke:
https://gist.github.com/coke/56f09da0b4793f9ec864 - here's a list of all the tickets that were opened, with their subject. I'll mark here what's been done to what until the patch is applied.

coke:
The list has been moved to https://github.com/perl6/mu/tree/master/misc/rt.perl.org so more than one person can work on it.

dnmfarrell:
Hey Will, I've fixed the quoting. See what you think!

usev6:
Hi, I've resolved the remaining conflicts and merged your PR. Thanks for working on this.

dnmfarrell:
woohoo, thanks!

PR #63, opened 2015-06-23
Adding a test for RT #125457 regressions

zostay:
This test fails on rakudo nom, nqp master, MoarVM master for the reasons I reported in RT #125457. Contributing this test to detect the error now and any future regressions.


dwarring:
@zostay - plan needs to be incremented: `plan 32;`

zostay:
Ah, oops... done!

PR #66, opened 2015-08-28
Create tests for the COMBINING GRAPHEME JOINER special case

alexghacker:
Tests for how the COMBINING GRAPHEME JOINER is handled under NFG.

Necessary because it is a special case of a combining character with ccc=0.

More extensive tests would include making sure that other combining characters don't move from one side of CGJ to the other, including "merging" with the base character through the selection of a precomposed canonically equivalent form under NFC.


zoffixznet:
Thanks a lot!

PR #70, opened 2015-09-19
Use 'run' to test for returning outside of routine

usev6:
We can't test for 'Attempt to return outside of any Routine'
with our toolchain from Test.pm or Test::Util.

### DIFF
--- a/S04-statements/return.t
+++ b/S04-statements/return.t
@@ -2,6 +2,9 @@ use v6;
 use Test;
 plan 20;
 
+use lib 't/spec/packages';
+use Test::Util;
+
 # Is there a better reference for the spec for how return return works? 
 # There is "return function" but that's a more advanced feature.
 #L<S04/"Control Exceptions">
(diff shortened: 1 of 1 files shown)


usev6:
I think the current tests don't work as intended.

```
$ perl6-m -e 'use Test; eval-dies-ok(q[return 1])'
ok 1 -
```

```
$ perl6-m -e 'use Test; eval-dies-ok(q[return])'
not ok 1 -
```

Compare this with 

```
$ perl6-m -e 'return 1'
Attempt to return outside of any Routine
  in block <unit> at -e:1
```

```
$ perl6-m -e 'return'
Attempt to return outside of any Routine
  in block <unit> at -e:1
```

Since even 'is_run' from Test::Util didn't give the
same results as running the code from the command line,
I resorted to 'run'.

This gives the expected results. On the downside it
makes the tests more expensive (esp. for rakudo.jvm). 
Furthermore I don't know if the tests run on Windows.

However, I decided not to push directly to roast, but
to create a pull request so that someone can review
the change. (Thanks in advance!)

PR #73, opened 2015-11-04
Add spectests for 1^..^10 non inclusive range quantifiers in regexs

ungrim97:
A recent change in NQP https://github.com/perl6/nqp/pull/249 added support for non inclusive range syntax as quantifiers in a regex. This adds spec tests to test that functionality


zoffixznet:
Thanks a lot!

PR #74, opened 2015-11-04
Test for RT #126270

molecules:
Please remove the CATCH block when RT #126270 is corrected.


molecules:
Going to simplify test and resubmit pull request

PR #75, opened 2015-11-05
expand S16-io/newline.t tests to include Buf decoding

ronaldxs:
Tested on both ubuntu linux and Windows 7


zoffixznet:
Thanks a lot!

PR #77, opened 2015-11-08
fix S17-supply/syntax.t race condition

LLFourn:
This fixes a race condition where 
```perl6
$lock.protect({ $cv1.signal; $cv2.wait });
```
would run before
```perl6
 $lock.protect({ $cv1.wait });
```
which causes the test to hang forever. It only happened when I had TEST_JOBS set to >= to my cores. Not sure if this is a symptom of a greater bug or just something the test writer didn't consider.

### DIFF
--- a/S06-operator-overloading/sub.t
+++ b/S06-operator-overloading/sub.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 93;
+plan 96;
 
 =begin pod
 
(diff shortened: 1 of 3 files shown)


LLFourn:
For posterity's sake:

I somehow discovered this erroneously named merge that I am responsible for.
This is because I didn't push to a clean branch or something. The actual merge was here: 

https://github.com/perl6/roast/pull/80

:-1: put it down to learning

PR #78, opened 2015-11-08
is required used as default checks

LLFourn:
RT #126436


LLFourn:
merge after: https://github.com/rakudo/rakudo/pull/579

PR #87, opened 2015-12-11
Fix Win32 specific IO::Spec tests so they pass

nanis:
I filed a confused [bug report][1] thinking these tests were not passing because of a failure of IO::Spec to delegate to IO::Spec::Win32. Later, I realized that invoking methods on IO::Spec is wrong, and one must use $*SPEC. This chunk of the the test file probably did not get exercised much.

[1]: https://rt.perl.org/Public/Bug/Display.html?id=126876


labster:
Looks good to me.  Invoking methods on IO::Spec wasn't wrong when I wrote the tests, but definitely became wrong later :)

nanis:
I figured as much. Thank you.

PR #88, opened 2015-12-17
Add test for subsecond ISO 8601 timestamp

Skarsnik:
see https://github.com/rakudo/rakudo/pull/638 and RT 126948


jonathanstowe:
The change has been merged now I think

PR #91, opened 2015-12-27
Add tests for advanced encode/decode

bluebear94:
See [here](https://gist.github.com/Juerd/ae574b87d40a66649692) for details.

Not sure where to put these tests, though.

### DIFF
--- a/S32-str/decode-advanced.t
+++ b/S32-str/decode-advanced.t
@@ -1,0 +1,132 @@
+use v6;
+use Test;
+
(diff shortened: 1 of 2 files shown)


JJ:
What spec do these tests cover?

AlexDaniel:
Sorry, so this is a proposal to change the language? Nowadays this should go through https://github.com/perl6/problem-solving/ first.

JJ:
Sorry, after 8 years I think it's best if we close this.

PR #94, opened 2016-01-05
Add tests for auto-generated private accessors.

autark:


AlexDaniel:
I am half way through reviewing this PR. However, I don't understand why the test for ``RT #125227`` is removed. Could you shine a light on that?

That being said, this removed test seems to be wrong. ``$o ~~ /':2'/`` will probably match ``World.nqp:2498`` just as well.

autark:
These tests are written in support of PR rakudo/rakudo#674. That PR adds support for auto-generating private accessors, it also attaches defined semantics to 'is rw'/'is readonly' for private attributes. In that regard the test was removed as it specifically tests that 'is rw' will generate a warning -- something it should no longer do.

JJ:
This issue mentions an issue that has been closed as Rejected, and also has a good amount of conflicts. Closing it.

AlexDaniel:
What? This is a PR for https://github.com/rakudo/rakudo/pull/674 which is open. So this should be open as well.

vrurg:
Worth a problem-solving issue, I'd say.

JJ:
Since https://github.com/rakudo/rakudo/pull/674 is closed, closing this one too. It's also conflicting, so it does not help anyone keeping it open.

PR #95, opened 2016-01-11
fix 455353534536453534563454353<[34]>

dankogai:
If you see carefully, messages like the following do not agree with the actual tests.

````perl6
# intentionally split with LF
is (-1) ** 4553535345364535345634543533, -1, 
    "-1 ** 4553535345364535345634543534 == -1";
````

This PR will fix that.

Dan the Perl6 Tester


AlexDaniel:
Right!

By the way, as far as I know some of these tests should cause numeric overflow now. Perhaps you could also take care of that?

PR #96, opened 2016-01-11
introduce X::Numeric::Overflow

dankogai:
Upon the request in https://github.com/perl6/roast/pull/95#issuecomment-170444410 .

### DIFF
--- a/S32-num/power.t
+++ b/S32-num/power.t
@@ -14,6 +14,7 @@ is(4 ** 2,   16, "4 **  2 == 16");
 
 my $big-e = 4553535345364535345634543534;
 my $big-o = 4553535345364535345634543533;
+my $xno = X::Numeric::Overflow;
 
 is 0 ** $big-e,     0, "0 ** $big-e == 0";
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Thank you for your pull request!

So, the X::Numeric::Overflow exception was added in https://github.com/rakudo/rakudo/commit/67795245fd9b17ca11036b63aa04e17deabb8e7a. It seems like that's the right behavior and that's how different implementations are supposed to behave now.

There is one thing that this pull request does not do though. These tests are fudged for moar, so if you are fixing them then I guess it would be right to unfudge them :)

Previously moar was returning **0** and jvm was returning **Inf**, so my guess is that's why these tests were fudged (0 is just blatantly wrong). That's before X::Numeric::Overflow was introduced.

Therefore I'll unfudge these tests on moar but fudge them on jvm shortly.

AlexDaniel:
I had to fudge these tests again because of OS X. See https://github.com/perl6/roast/commit/2ff9b0910898066caf2c394502e73b7080d779f8

PR #99, opened 2016-01-19
Test hyper and race in reverse

b2gills:
Notice that the `.race.map` is the only one that currently works.


AlexDaniel:
So I reviewed the tests and I think that this pull request is great. Unlike my other attempt these tests will fail/pass consistently.

I find these tests a little bit too verbose (and a little bit too copy-pasted) but sure enough we can golf them later.

Thank you for your work.

PR #100, opened 2016-01-27
Test for require file where symbol isn't in module

LLFourn:
merge when this is https://github.com/rakudo/rakudo/pull/694 is merged


LLFourn:
https://github.com/perl6/roast/commit/25c40d05f70ba3fabb83cb35a714344bc34c449f

PR #101, opened 2016-02-06
Tests that Rat.perl to actually generates valid code

autarch:
See https://github.com/rakudo/rakudo/pull/706 for the corresponding rakudo PR to actually fix this.


autarch:
See https://github.com/rakudo/rakudo/pull/707 for a second attempt to fix this in rakudo.

PR #102, opened 2016-02-07
Tests for ｢hyper｣

AlexDaniel:
These tests stand out from the rest a little bit. Bug described in #127452 happens very frequently, but sometimes the code works. Is there any other way to test that without running the same code 10 times?

Same with ``^1000 .hyper.map: * + 10``. We can't know if it will actually mangle the result, but most probably it will.

### DIFF
--- a/S07-hyperrace/hyper.t
+++ b/S07-hyperrace/hyper.t
@@ -1,6 +1,6 @@
 use Test;
 
-plan 11;
+plan 12;
 
 {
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Ah! I've just noticed https://github.com/perl6/roast/pull/99. It seems like it tests the same thing in a much nicer manner.

AlexDaniel:
Yep.

PR #103, opened 2016-02-07
Add a test for the round trip of Instant via .perl.EVAL

autarch:


autarch:
Fixed in https://github.com/rakudo/rakudo/pull/708

PR #104, opened 2016-02-12
Fix tests that rely on Exception.handled

b2gills:
Now that you can set `Exception.handled` let's actually test it.

I changed the test to assign `True` rather than `1` so that in the future it could be implemented with `has bool $.handled is rw;` rather than as a proxied rw method which it is currently.

While looking into this I found two tests that were looking at `$!.pending` even though `$!` should be empty at that point. I think it was just a copy paste error. I changed one to look at `@fails` instead, and removed the other one.

Since all of the tests I changed were skipped or broken anyway, there should be no problems with maintaining compatibility with v**6.c**.

---

Actually thinking about it `.handled` might be better if it was called `.is-handled` as someone could think that just calling `.handled` causes it to be handled.

It should be safe to do this as the method was never there to begin with.

### DIFF
--- a/S04-exceptions/pending.t
+++ b/S04-exceptions/pending.t
@@ -1,6 +1,6 @@
 use v6;
 use Test;
-plan 21;
+plan 23;
 
 # XXX I'm not very confident in my reading of S04, so give a suspicious eye
 #     to these tests before using them.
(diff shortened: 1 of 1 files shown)


salortiz:
"S04/Exceptions" talks about `@!` as an unthrown exception list of Failures, so my though was that this S04-exceptions/pending.t tests was, in principle, about them.

I assumed that if `$! !~~ Exception`, then it should be `Failure`, and `.pending` the planned way to access that list. (Currently `$!` is `Exception` only after a failed `try` block)

That way, all tests made sense.

About `.handled,` in S04 it is described as a method of `Mu` only overridden by `Failure` and I start working on it when I found it described in standard documentation, see https://github.com/perl6/doc/blob/master/doc/Type/Failure.pod#L31, try to use it and failed.

Its evident that S04 and the current implementation are diverging, so I take roast as normative, and this PR removes that way for testing pending Failures. I that isn't desired I haven't drawbacks.

(`Mu.handled` is trivial, but I don't see its case.)

JJ:
Looks like this is still needed, but can someone please check the conflicts?

JJ:
... No one did check the conflicts, so closing too.

PR #110, opened 2016-04-01
More tests for combinations, including pathological cases

skids:


skids:
To be merged after rakudo PR#732

moritz:
Why do you compare .perl output in those tests, instead of just is-deeply? This looks like an anti-pattern to me.

skids:
Mainly just because that is what the pre-existing code did.  Might predate is-deeply.

zoffixznet:
Thanks. There were merge conflicts so I manually added these in https://github.com/perl6/roast/commit/1424793070962b78a9b3d366a808cb89ed27b1ed

PR #112, opened 2016-04-13
Duration Improvements

pprince:
This is my first attempt to contribute something to p6; feel free to point out any errors/misunderstandings on my part, or further improvements; I am happy to incorporate suggestions and re-submit.

This adds a lot of test cases; in the long term, I can pare them down to the essentials, but for the moment and while the Duration behavior/implementation is in flux, i found them useful.

Description:
---
    Add additional tests for Duration; see RT #127339
    
    Also see corresponding PR/commits against rakudo/rakudo.
    
    Has tests to cover new functionality, and all tests pass.
    
    Still, I am new to Perl6, and look forward to one or more
    rounds of requested changes before this gets into Perl6.
---
    Move Duration and Instant closer to S02 spec.
    
    Duration and Instant should behave like value types.
    
    Also, Duration should become a Rat in numeric context.
---
    Fix warning and add a case for Instant round-trip
    
    Removes warning due to redefining "my $i".
    
    Adds Instant.from-posix(915148800) to the test values; this should
    mean 1998-12-31T23:59:60Z when $prefer-leap-second is False, but
    mean 1998-12-31T23:59:60Z 1999-01-01T00:00:00Z when it is True.

### DIFF
--- a/S02-types/instants-and-durations.t
+++ b/S02-types/instants-and-durations.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 51;
+plan 113;
 
 # L<S02/Immutable types/'term now'>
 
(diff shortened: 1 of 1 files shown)


pprince:
**SEE ALSO:  [rakudo/rakudo PR#742](https://github.com/rakudo/rakudo/pull/742)**

pprince:
I am going to withdraw this pair of PR's for now, as I:

1. have re-considered the ideal corner-case behavior,
2. would like to submit cleaner PR's that don't e.g. leave commit-squashing up to the person merging,
3. would like to not litter roast with extraneous test cases, and
3. would like to simultaneously offer a PR to update/clarify the Duration-related parts of S02 & S32.

You should look forward to my re-submssion... I believe we can specify Duration such that everyone gets what they want, no one gets any surprises, "weird" cases Just Work™, and all while avoiding having to do anything you'd call complicated.  Wish me luck :)

PR #115, opened 2016-04-21
Pod on declarator with initialization

bduggan:
Hi All,

The attached tests are failing for me with the most recent rakudo.

My reading of S26 would be that it should be valid to have declarator pod trailing an attribute even if there is also an initialization of the attribute. 

thanks
Brian

### DIFF
--- a/S26-documentation/block-trailing.t
+++ b/S26-documentation/block-trailing.t
@@ -1,6 +1,6 @@
 use v6;
 use Test;
-plan 327;
+plan 334;
 
 my $pod_index = 0;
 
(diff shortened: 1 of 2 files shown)


MadcapJake:
IIRC, the `#=` form doesn't require any enclosing brackets.

bduggan:
Right...they're optional -- block-trailing.t has tests with enclosing brackets, and why-trailing.t has tests without enclosing brackets.

moritz:
Thanks @bduggan. I've merged it, converted the tests to use subtests, and fudged the new, failing tests for rakudo.

PR #120, opened 2016-05-14
Add from-loop tests

### LINKED ISSUE #114: from-loop is not tested
If you want to know what from-loop is, see http://doc.perl6.org/routine/from-loop

titsuki:
Fix #114

I have added tests for the from-loop method according to the following page
https://doc.perl6.org/routine/from-loop#class_Seq

But I don't have very much confidence. Can anyone review this tests ?

The following is the results of the from-loop tests
```
not ok 29 - from-loop(&body) returns a Seq

# Failed test 'from-loop(&body) returns a Seq'
# at ./S32-list/seq.t line 133
ok 30 - the Seq object is lazy
not ok 31 - from-loop(&body, &condition) returns a Seq

# Failed test 'from-loop(&body, &condition) returns a Seq'
# at ./S32-list/seq.t line 137
ok 32 - from-loop(&body, &condition) terminates calling &body if &condition returns False
not ok 33 - from-loop(&body, &condition, &afterward) returns a Seq

# Failed test 'from-loop(&body, &condition, &afterward) returns a Seq'
# at ./S32-list/seq.t line 143
ok 34 - &afterward is called after each call to &body.
ok 35 - from-loop(&body, &condition, &afterward) terminates calling &body if &condition returns False
```

### DIFF
--- a/S32-list/seq.t
+++ b/S32-list/seq.t
@@ -128,19 +128,17 @@ is-deeply @searches[0].Array, @expected-searches, 'seq => array works 3';
 }
 
 {
-    ok Seq.from-loop({ 1 }).WHAT === Seq, 'from-loop(&body) returns a Seq';
-    my @a;
-    @a = Seq.from-loop({ 1 });
-    is @a.is-lazy, True, 'the Seq object is lazy';
+    my $a = Seq.from-loop({ 1 });
+    isa-ok $a, Seq, 'from-loop(&body) returns a Seq';
+    is $a.is-lazy, True, 'the Seq object is lazy';
 
-    ok Seq.from-loop({ 1 }, { state $count = 0; $count++ < 10 }).WHAT === Seq, 'from-loop(&body, &condition) returns a Seq';
-    @a = Seq.from-loop({ 1 }, { state $count = 0; $count++ < 10 });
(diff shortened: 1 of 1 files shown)


titsuki:
Sorry, I have fixed the way to check returned type.
And the following is the results
```
ok 28 - .perl on an iterated sequence faithfully reproduces such a sequence
ok 29 - from-loop(&body) returns a Seq
ok 30 - the Seq object is lazy
ok 31 - from-loop(&body, &condition) returns a Seq
ok 32 - from-loop(&body, &condition) terminates calling &body if &condition returns False
ok 33 - from-loop(&body, &condition, &afterward) returns a Seq
ok 34 - &afterward is called after each call to &body.
ok 35 - from-loop(&body, &condition, &afterward) terminates calling &body if &condition returns False
```

salortiz:
@titsuki 
The type of an object can be tested by 
```perl6
isa-ok Seq.from-loop({ 1 }), Seq, "Bla, bla";
```
and you can avoid the coercion to `Array` using a `Scalar` variable (or sigilless):
```perl6
     my $foo = Seq.from-loop({ 1 });
     ok $foo.is-lazy, …
```

titsuki:
@salortiz 
Thank you for your advice !
I have fixed your pointer.

PR #121, opened 2016-05-26
add critical test for known critical table pod bug (#128221) in lates…

tbrowder:
…t rakudo release (2016.05)

### DIFF
--- a/S26-documentation/07-tables.t
+++ b/S26-documentation/07-tables.t
@@ -142,5 +142,5 @@ is $r.contents.elems, 2;
 =end table
 
 $r = $=pod[9];
-is $r.contents.elems, 2;
+is $r.contents.elems, ;
 
--- a/S26-documentation/07-tables.t
+++ b/S26-documentation/07-tables.t
@@ -1,6 +1,6 @@
 use v6;
 use Test;
-plan 38;
+plan 39;
 my $r;
 
 =begin table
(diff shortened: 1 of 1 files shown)


tbrowder:
Note that IMHO this should be a release stopper until the test passes.

moritz:
@tbrowder is it a regression? If not, it's not a release stopper.

tbrowder:
On Tue, May 31, 2016 at 3:38 PM, Moritz Lenz <notifications@github.com> wrote:
> @tbrowder is it a regression? If not, it's not a release stopper.

Moritz, you're right I guess, but it seems to be a serious Perl 6 bug
that's been there that no one had found.

Do I need to change or revert the pull merge, or something else?

I would love to try to fix the bug but have not made much progress and
probably won't unless someone can help point the way.

Anyway, let me know what I need to do.  I do feel the test needs to
remain, but maybe not as a show-stopper test.

-Tom

jonathanstowe:
The test can be skipped or todo'd pending a fix:)

tbrowder:
On Tue, May 31, 2016 at 4:27 PM, Jonathan Stowe
<notifications@github.com> wrote:
> The test can be skipped or todo'd pending a fix:)

The 'todo' still results in an exception due to the bug, so I
commented it out with what I hope is sufficient info for the future.
I'll remove the whole mess if that is preferred.

PR #123, opened 2016-06-02
Add some tests for the ≅ operator

MasterDuke17:


AlexDaniel:
What about these? https://github.com/perl6/roast/blob/master/S03-operators/relational.t#L211

MasterDuke17:
Ha. Just looked back in my history and I only searched for ≅ in S32-num, not the whole suite. Think I should just move my tests to alongside the others in S03-operators/relational.t and not create a new file in S32-num?

AlexDaniel:
Yeah. I think that the ≅ alone does not justify the creation of a new file. But it does not really matter which way you go, as long as similar tests are kept in one place.

PR #136, opened 2016-07-20
add and modify files to test expected exceptions for known illegal Pe…

tbrowder:
…rl 6 pod tables


tbrowder:
trying suggestions from Zoffix...will retry later

PR #137, opened 2016-08-02
Reading /proc files line by line does not go into infinitive loop

Gnouc:
RT #127370


Gnouc:
PR associated with this test was merged https://github.com/MoarVM/MoarVM/pull/388

PR #146, opened 2016-08-24
Tests for #126124

LemonBoy:


S06-signature/types.t
```diff
@@ -35,4 +35,14 @@ dies-ok  { g('m') },    'type check forbids bad implicit return';
         'no segfault when calling a routine having a native parameter with a type object argument';
 }
 
+# RT #126124
+{
+    throws-like { sub f(Mu:D $a) {}; f(Int) }, Exception,
+        message => 'Parameter \'$a\' requires an instance of type Mu, but a type object was passed.  Did you forget a .new?',
+        'type shown in the exception message is the right one';
+    throws-like { sub f(Mu:U $a) {}; f(123) }, Exception,
+        message => 'Parameter \'$a\' requires an instance of type Mu, but a type object was passed.  Did you forget a .new?',
```
zoffixznet: Seems this line expects the wrong message. If the test is passing, something wrong, because the message should contain `but an object instance was passed`, yet the test tests for `but a type object was passed`

LemonBoy: oops, copy&paste error. (force) pushed a working version!

jnthn: Hmm, do we not have a typed exception for this? Testing against precise messages is a bit fragile.

zoffixznet:
Merged. Will do a commit to check for message with a shorter regex, to make it less flaky, [as per IRC conversation](http://irclog.perlgeek.de/perl6-dev/2016-08-24#i_13081530).

PR #151, opened 2016-09-05
Tests for RT #128035

MasterDuke17:
Large repetition values for Str.x should not complain about negative
numbers or give an empty string as the result.

Requires Rakudo PR #867 (https://github.com/rakudo/rakudo/pull/867).

### DIFF
--- a/S03-operators/repeat.t
+++ b/S03-operators/repeat.t
@@ -8,7 +8,7 @@ Repeat operators for strings and lists
 
 =end description
 
-plan 44;
+plan 48;
 
 #L<S03/Changes to Perl 5 operators/"x (which concatenates repetitions of a string to produce a single string">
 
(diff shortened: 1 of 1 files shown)


S03-operators/repeat.t
```diff
@@ -44,6 +44,24 @@ is 'str' x Int, '', 'x with Int type object';
...
+        X::AdHoc,
+        message => 'repeat count > 1073741824 arbitrarily unsupported...',
+        'repeat count greater than the NQP limit dies'
+    );
+    throws-like(
+        { 'a' x int.Range.max + 1 },
+        X::AdHoc,
+        message => 'repeat count > 1073741824 arbitrarily unsupported...',
+        'repeat count greater than the max size of an int dies because of the NQP limit'
+    );
```
moritz: This last test fails on current Rakudo, because `int.Range.max + 1` evaluates to 0.

PR #155, opened 2016-09-06
Make sure where-blocked multi is prior to non-where-blocked multi

### LINKED ISSUE #154: Tie-breaking rule between where-blocked multi and non-where-blocked multi is not specced
I can find the tie-breaking rule between **where-blocked multi** routines in roast.
https://github.com/perl6/roast/blob/master/S06-multi/subsignature.t#L431-L441

However, it seems that there is another tie-breaking rule that **where-blocked multi** has higher priority than **non-where-blocked multi**.

The following example shows the above mentioned rule.
```
$ perl6 -e 'multi foo(Int $a){ say "1st" }; multi foo(Int $a where *.so){ say "2nd" }; foo(-1)'
2nd
```

```
$ perl6 -e 'multi foo(Int $a where *.so){ say "1st" }; multi foo(Int $a){ say "2nd" }; foo(-1)'
1st
```

titsuki:
Fix #154

### DIFF
--- a/S06-multi/syntax.t
+++ b/S06-multi/syntax.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 44;
+plan 45;
 
 # L<S06/Routine modifiers/>
 # L<S06/Parameters and arguments/>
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks.

Issue #166, opened 2016-10-05
no test for Proc.command and Proc.signal


gfldex:

gfldex:
please close https://github.com/perl6/doc/issues/913 when the test is added

PR #169, opened 2016-10-15
Tests for Proc.command and Proc.signal

hchienjo:
Changes
===========
Add tests for `.command` and `.signal` methods for proc.

Proposed fix for https://github.com/perl6/roast/issues/166

### DIFF
--- a/S29-os/system.t
+++ b/S29-os/system.t
@@ -23,7 +23,7 @@ $res = shell("$*EXECUTABLE -e \"\"");
 ok($res, "shell() to an existing program does not die (and returns something true)");
 isa-ok($res, Proc, 'shell() returns a Proc');
 is($res.exitcode, 0, 'shell() exit code when successful is zero');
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks a lot!

To note as far as test writing goes, the test for `.signal` is incomplete, as it does not test cases where `.signal` is *not* 0. In fact, there's actually a bug with it!

There's an [open ticket for that bug](https://rt.perl.org/Ticket/Display.html?id=128805#ticket-history) and the test would require [fudging](https://github.com/perl6/roast/blob/master/README#L22). Here's [an example](https://github.com/perl6/roast/blob/master/S06-signature/code.t#L48-L50) of what a fudged test looks like. If you're up for that, perhaps there's another PR you can make to count for the Hacktoberfest? 😄

hchienjo:
@zoffixznet That is what I have been asking on the IRC today. Wanted to test for other cases apart from 0. Will look into it.

PR #173, opened 2016-10-19
A couple more duckmap tests

cpin:
Assuming I have read between the lines of http://irclog.perlgeek.de/perl6/2016-09-26#i_13285724 correctly, these should verify `duckmap` behaviour slightly more. Comments?

### DIFF
--- a/S32-list/duckmap.t
+++ b/S32-list/duckmap.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 2;
+plan 5;
 
 =begin description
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
More tests are good, but you've read between the lines wrong. The issue is this:

```
<Zoffix__> m: use Test; my $list = (1, [2,3], "a"); dd duckmap(-> Int $x { $x ~~ Int ?? $x + 1 !! Any }, $list)
<camelia> rakudo-moar eb6907: OUTPUT«(2, (3, 4), "a")␤»
```

Note how it changed `[...]` inner array to `(...)` inner list, but it must keep it as an array.

cpin:
Yes, I realise that - I added tests for the documented behaviour as it seems there was some confusion over behaviour that was not related to structural conservation. So this isn't intended to provide tests for RT #129363 (a case already exists, and is skipped for Rakudo), but rather to document the existing behaviour.

PR #175, opened 2016-10-20
Test using fail in subset for custom message

zostay:


zoffixznet:
Would it be possible to change this test to check the actual content of the custom message is correct? The current `dies-ok` version takes the `1 is not even` string as test description; not the content of the death to check against. For those purposes, there's [`throws-like`](https://docs.perl6.org/language/testing#index-entry-throws-like):

```perl6
throws-like { fail "foo" }, Exception, :message<foo>,
    'failure gave correct custom message';
```

zostay:
Oh... duh... yeah, I'll fix it.

zoffixznet:
Also, you forgot to adjust the `plan` at the top of the file :smile_cat:

zostay:
Fixed the plan and using `throws-like` now. Thx for the review.

zoffixznet:
Thanks a lot.

PR #179, opened 2016-11-02
Test updated for proper bag/mix symmetric difference

ab5tract:
These are the tests for PR https://github.com/rakudo/rakudo/pull/911

### DIFF
--- a/S03-operators/mix.t
+++ b/S03-operators/mix.t
@@ -66,18 +66,16 @@ sub symmetric-difference($a, $m) {
     ($a (|) $m) (-) ($m (&) $a)
 }
 
-#?rakudo 8 todo "Rakudo update in progress, but not done yet RT #124541"
-
-is showkv($b (^) $m), showkv(symmetric-difference($b, $m)), "Mix symmetric difference with Bag is correct";
+is ($b (^) $m), symmetric-difference($b, $m), "Mix symmetric difference with Bag is correct";
 isa-ok ($b (^) $m), Mix, "... and it's actually a Mix";
-is showkv($m (^) $b), showkv(symmetric-difference($b, $m)), "Bag symmetric difference with Mix is correct";
+is showkv($m (^) $b), symmetric-difference($b, $m), "Bag symmetric difference with Mix is correct";
 isa-ok ($m (^) $b), Mix, "... and it's actually a Mix";
 
 #?niecza todo "Test is wrong, implementation is wrong"
-is showkv($b (^) $mh), showkv(symmetric-difference($b, $mh)), "MixHash symmetric difference with Bag is correct";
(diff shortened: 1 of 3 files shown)


S03-operators/mix.t
```diff
@@ -217,19 +211,14 @@ ok mix(my @large_arr = ("a"...*)[^50000]), "... a large array goes into a bar -
...
-    is showkv([(^)] $b, $m, $mh), showkv({ blood => 1, love => 1, rhetoric => 1 }), "Mix symmetric difference reduce works on three mixs";
-    isa-ok showkv([(^)] $b, $m, $mh), Mix, "Mix symmetric difference reduce works on three mixs";
+    is ([(^)] $m), $m, "Mix symmetric difference reduce works on one mix";
+    isa-ok ([(^)] $m), Mix, "Mix symmetric difference reduce works on one mix, yields mix";
+    is ([(^)] $b, $m), (blood => 0.1, love => 0.2, :rhetoric).Mix, "Mix symmetric difference reduce works on a mix and a set";
+    isa-ok ([(^)] $b, $m), Mix, "... and produces a Mix";
+    is ([(^)] $m, $b), (blood => 0.1, love => 0.2, :rhetoric).Mix, "... and is actually symmetric";
+    isa-ok ([(^)] $m, $b), Mix, "... and still produces a Mix that way too";
+    is ([(^)] $b, $m, $mh), (blood => 1, love => 1.1, :rhetoric).Mix, "Mix symmetric difference reduce works on three mixs";
+    isa-ok ([(^)] $b, $m, $mh), Mix, "Mix symmetric difference reduce works on three mixs";
```
zoffixznet: Are you positive this test file still passes? I see 10 tests in deleted lines, but only 8 in the added lines. Does the plan need adjustment?

ab5tract: Ah yes, these were two redundant tests of set/bag combinations.

zoffixznet:
These will also need to be replicated in the `6.c-errata` branch. As they were all TODOed and didn't work right, I don't consider this violating the 6.c terms.

S03-operators/mix.t
zoffixznet: FWIW, you can run `make t/spec/S03-operators/mix.t` from a built Rakudo repo and it'll run just that test file with the harness and fudge.

zoffixznet:
Thanks a lot.

I don't think we need to propagate these to 6.c-errata; those were all fudged anyway and there's no point in altering the tests.

zoffixznet:
I had to [revert](https://github.com/perl6/roast/commit/0c09883bef4a61eca105e0c6eb17a0a4a37fc9ac) this and the associated Rakudo PR. See [comments on Rakudo's PR](https://github.com/rakudo/rakudo/pull/911#issuecomment-258854543) for explanation.

Issue #161, opened 2016-09-27
.duckmap/.deepmap are entirely untested.


zoffixznet:
See http://irclog.perlgeek.de/perl6-dev/2016-09-27#i_13295357

There's a passing mention in `S03-metaops/hyper.t` to test them for nodalness, but the functionality itself isn't tested.

---
### Hacktoberfest participants:

The behaviour can be found from sourcecode [here](https://github.com/rakudo/rakudo/blob/222d16b/src/core/metaops.pm#L683) and [here](https://github.com/rakudo/rakudo/blob/222d16b/src/core/metaops.pm#L553). the docs [here](https://docs.perl6.org/routine/deepmap) and [here](https://docs.perl6.org/routine/duckmap), or you can join our IRC channels and ask for clarifications [#perl6](https://webchat.freenode.net/?channels=#perl6) or [#perl6-dev](https://webchat.freenode.net/?channels=#perl6-dev).

Create new test files: `S32-list/duckmap.t` (might already exist) and `S32-list/deepmap.t`. Copy over one of the existing files in that directory to use as a starting template. Use any of [the available testing routines](https://docs.perl6.org/language/testing.html#Reference) to thoroughly test all of the behaviour and then submit the PR.

Once done and the PR is merged, if you _did_ create new files, you can also submit another PR to [add them to Rakudo's test list](https://github.com/rakudo/rakudo/blob/nom/t/spectest.data#L974); just add your file where it fits alphabetically.

PR #181, opened 2016-11-10
Add initial tests for deepmap (#161)

cpin:


zoffixznet:
Thanks a lot. As this is a new test file, it also needs to be added [in Rakudo's t/spectest.data file](https://github.com/rakudo/rakudo/blob/nom/t/spectest.data#L984-L985). Would you, please, submit a PR adding it?

cpin:
Certainly will - in fact, am just in process of doing so.

PR #182, opened 2016-11-14
Add tests for Qw, Qx and Qww. One more test for qx to check how two backslashes are interpreted

samcv:


S02-literals/quoting.t
```diff
@@ -524,6 +527,8 @@ Hello, World
     my $var = 'world';
     is  qx/echo world/.chomp, "world", 'qx';
     is qqx/echo $var/.chomp,  "world", 'qqx';
+    is Qx[echo '\\'].chomp, "\\\\", "Qx treats backslash literally";
```
zoffixznet: This test fails. With different failure modes on Debian and on Windows:

**Debian:**
```
not ok 1 - Qx treats backslash literally

# Failed test 'Qx treats backslash literally'
# at z.p6 line 4
# expected: '\\'
#      got: '\'
```

**Windows:**
```
not ok 1 - Qx treats backslash literally

# Failed test 'Qx treats backslash literally'
# at p.t line 4
# expected: '\\'
#      got: ''\\''
```

S02-literals/quoting.t
```diff
@@ -524,6 +527,8 @@ Hello, World
...
+    is qx[echo '\\'].chomp, "\\", 'qx escaping a backslash returns one backslash';
```
zoffixznet: This test fails on Windows:
```
# Failed test 'qx escaping a backslash returns one backslash'
# at p.t line 5
# expected: '\'
#      got: ''\''
```

samcv:
It should pass now I think on Windows and POSIX environments. Since echo on Windows leaves quotation marks in, and on POSIX we need them to treat the contents literally, compare qx[echo '\\\\\\\\'] and Qx[echo '\\\\'] to have the same result. The reason for so many slashes is this:
**bash, sh, dash (echo binary) and csh builtin:**
`echo '\' #> \`
`echo '\\' #> \\`
**zsh builtin and tcsh builtin:**
`echo '\' #> \`
`echo '\\' #> \`
**Windows:**
`echo '\' #> '\'`
`echo '\\' #> '\\'`

zsh and tcsh's builtin echo does not properly follow POSIX. Although worst case would be that the test would pass when it shouldn't, we can do better than this. We fix this with the extra slashes. As you can see all the below are different, and the test is functional on POSIX compliant and noncompliant echo's
**zsh builtin and tcsh builtin:**
`echo '\\\\\\\\' #> \\\\`
`echo '\\\\' #> \\`
`echo '\\' #> \`

If something went wrong and Qx may send half as many slashes as expected, this should detect it without assuming which shell the user has set as the system shell.

zoffixznet:
Thanks a lot!

PR #183, opened 2016-11-18
Make test for output of backtrace less specific

usev6:
After the recent changes to backtrace handling this test
started to fail on rakudo-j. There are references to lines
in gen/jvm/CORE.setting which are causing the test to fail
because of the 's' in 'setting'.

Allowing additional lines in the backtrace still fits the test
description 'Thrown Failure outputs dual backtraces'.

### DIFF
--- a/integration/error-reporting.t
+++ b/integration/error-reporting.t
@@ -167,7 +167,7 @@ ok $b1 === $b2, "Backtrace does not change on additional .backtrace";
 }
 
 is_run 'sub s1 { sub s2 { fail("foo"); }; s2()(); }; s1();', {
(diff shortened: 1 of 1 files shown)


usev6:
```
$ ./perl6-j -e 'sub s1 { sub s2 { fail("foo"); }; s2()(); }; s1();'
foo
  in sub s2 at -e line 1
  in sub s1 at -e line 1
  in block <unit> at -e line 1

Actually thrown at:
  in method throw at gen/jvm/CORE.setting line 26819
  in method throw at gen/jvm/CORE.setting line 29478
  in method CALL-ME at gen/jvm/CORE.setting line 29529
  in sub s1 at -e line 1
  in block <unit> at -e line 1
```

PR #185, opened 2016-11-28
Fix test which breaks on 32 bit systems

dogbert17:
The test for RT #112788 didn't work properly on 32-bit systems. This
has been fixed by changing the exponent from 9 to 8 so now it should
behave the same, i.e. hang, on both 32 and 64 bit systems. An
additional test was added (viki++) which checks that a Rat raised to
a very large number actually throws.


zoffixznet:
Thanks.

PR #187, opened 2016-12-04
Re-introduce the fixes for bag/mix symdif

ab5tract:
This time, we don't change any of the set tests.

All of the 'showkv' calls were removed as they are no longer necessary
and several new tests to ensure symmetric behavior on (+) and (.) operators.

### DIFF
--- a/S03-operators/bag.t
+++ b/S03-operators/bag.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 128;
+plan 130;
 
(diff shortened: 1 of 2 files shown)


zoffixznet:
How come `showkv` are no longer needed? Setties/Baggies are unordered so the how they stringify is not predicable, or is it?

I know if the tests are changed to `is-deeply` instead of `is` then we won't need showkv for sure....

PR #189, opened 2016-12-08
A test for #130291

Altai-man:
I wonder if this test good enough, feel free to edit or point out anything.

It checks stdout and stderr. It fails on rakudo before a fix and passes on latest.

### DIFF
--- a/S17-procasync/basic.t
+++ b/S17-procasync/basic.t
@@ -107,7 +107,7 @@ throws-like { Proc::Async.new }, X::Multi::NoMatch,
         "Tapping stdout supply after start of process does not lose data";
 }
 
-$pc = Proc::Async.new("perl6", "-e ''"); # It taps a warning, but it is safe to catch.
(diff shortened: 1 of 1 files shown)


S17-procasync/basic.t
```diff
@@ -106,3 +106,12 @@ throws-like { Proc::Async.new }, X::Multi::NoMatch,
     is $captured, "Hello World\n",
         "Tapping stdout supply after start of process does not lose data";
 }
+
+$pc = Proc::Async.new("perl6", "-e ''"); # It taps a warning, but it is safe to catch.
```
zoffixznet: Would you change this to `$*EXECUTABLE` instead of `"perl6"`?

That way we don't rely on there being an executable named perl6.

Altai-man: Sure! I didn't know what command is better to write, since things like "ls" are quite UNIX-like-specific.

zoffixznet:
Thanks a lot!

PR #192, opened 2016-12-10
Test for #125247

Altai-man:
https://rt.perl.org/Public/Bug/Display.html?id=125247#ticket-history

The original ticket is about a proper error message for an unfinished multi-line comment. The error message should contain right line number, i.e. where the comment started.


zoffixznet:
Thanks a lot!

PR #193, opened 2016-12-11
Add tests for uniprops and add a fudged test for a MoarVM bug

samcv:
https://github.com/MoarVM/MoarVM/issues/448

### DIFF
--- a/S15-unicode-information/uniprop.t
+++ b/S15-unicode-information/uniprop.t
@@ -19,7 +19,7 @@ throws-like "Int.uniprop", X::Multi::NoMatch, 'cannot call uniprop with a Int';
 is "".uniprops, (), "''.uniprops yields an empty list";
 is ("\x[1000]", "\x[100]")».uniprop('Script'), "\x[1000]\x[100]".uniprops('Script'),
     "uniprops returns properties of multiple characters in a string";
-#?rakudo.moar skip "Unicode properties don't always return the correct shortnames"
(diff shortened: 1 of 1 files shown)


S15-unicode-information/uniprop.t
```diff
@@ -2,18 +2,25 @@ use v6;
...
 is "".uniprop, Nil, "''.uniprop yields Nil";
 throws-like "uniprop Str", X::Multi::NoMatch, 'cannot call uniprop with a Str';
 throws-like "Str.uniprop", X::Multi::NoMatch, 'cannot call uniprop with a Str';
 throws-like "uniprop Int", X::Multi::NoMatch, 'cannot call uniprop with a Int';
 throws-like "Int.uniprop", X::Multi::NoMatch, 'cannot call uniprop with a Int';
 
+is "".uniprops, (), "''.uniprops yields an empty list";
+is ("\x[1000]", "\x[100]")».uniprop('Script'), "\x[1000]\x[100]".uniprops('Script'),
+    "uniprops returns properties of multiple characters in a string";
+#?rakudo.moar skip "Unicode properties don't always return the correct shortnames"
```
zoffixznet: Would you, please, change this fudge to `todo`. This way the test will actually run and the harness will tell us when it starts to pass (in case the problem is fixed without knowing of that issue or about a fudged test).

```
#?rakudo.moar todo "Unicode properties don't always return the correct shortnames"
```

samcv:
Done.

zoffixznet:
Thanks a lot!

PR #196, opened 2016-12-15
A test for #130211

Altai-man:
Related to https://github.com/rakudo/rakudo/pull/957. Closes https://rt.perl.org/Public/Bug/Display.html?id=130211.

I just wrote checking of reverse-ordered role names to be independent from exact message.

### DIFF
--- a/integration/error-reporting.t
+++ b/integration/error-reporting.t
@@ -217,9 +217,10 @@ is_run Q[#`{{ my long
 	      { :out(''), :err{ $^o.contains: 'line 1' }}, 'Unfinished comment error points on correct line';
 
 # RT 130211
-throws-like 'role R { method overload-this(){...} };
-             role C { method overload-this(){...} };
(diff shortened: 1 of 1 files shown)


zoffixznet:
This test is a bit flimsy, since regular words in the message may contain letters 'C' and 'R'. Also, I don't think the Perl 6 Language should dictate in which order implementations must list the roles in this message.

I'd suggest you name the roles something more specific, like `R-RT130211` and `C-RT130211`, and simply check for presence of both in the message, instead of testing a specific order, so `message` would be something like:

```perl6
:message{ .contains('R-RT130211') and .contains('C-RT130211') }
```

Altai-man:
Indeed, your approach is much better.

zoffixznet:
> I get 255 error on my machine, so don't merge it yet.

That's because you put a space between `:message` and `{`

Altai-man:
I'll squash latest commits now...

zoffixznet:
Thanks a lot.

Altai-man:
Oh, I'm late. Never mind then.

Issue #200, opened 2016-12-22
S03-operators/precedence.t could be improved with ⁿ unicode superscript exponentiation tests


ronaldxs:
Unicode superscript exponentiation (ⁿ) tests were added to [S32-num/power.t](https://github.com/perl6/roast/blob/master/S32-num/power.t) but related tests did not seem to be added to [S03-operators/precedence.t](https://github.com/perl6/roast/blob/master/S03-operators/precedence.t) which also has tests for the ** operator.  Some tests like `++$i ** 2` (now `++$i²`) would not behave the same way because the superscript operator is not associative, but other tests should be quite similar.  Fixing this might add about a half dozen tests.

zoffixznet:
IMO this may be a bug. I'd expect `++$i²` to be exactly equivalent to `++$i ** 2`

zoffixznet:
TimToady's response: https://irclog.perlgeek.de/perl6-dev/2016-12-22#i_13785998

ronaldxs:
Just to clarify:

`2² ** 3` should be one of the test cases and it currently evaluates to 64 rather than `2 ** 2 ** 3 == 256`

zoffixznet:
Same cause as the difference in `++$i` case.

ronaldxs:
First cut at roast tests and opportunity to comment before PR. 

Tests on  [lines 216 through 227](https://github.com/ronaldxs/roast/blob/4c5541f2e1a225d668bc03d077d998510068d1e9/S03-operators/precedence.t#L216)

zoffixznet:
Re-opening this so we'd have a reference... This stuff still needs implementing.

PR #205, opened 2016-12-28
Superscript exponent precedence tests

ronaldxs:
Attempt to resolve Issue #200

### DIFF
--- a/S03-operators/precedence.t
+++ b/S03-operators/precedence.t
@@ -21,16 +21,18 @@ plan 77;
 # FIXME how do we test this?
 
 # postfix method
-
-my @a = 1,2,3;
-is(++@a[2], 4, "bare postfix binds tighter than ++");
-is(++@a.[2], 5, "dotted postfix binds tighter than ++");
+{
+    my @a = 1,2,3;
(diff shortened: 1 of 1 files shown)


S03-operators/precedence.t
```diff
@@ -213,6 +213,19 @@ is(@c, [1,2,3], "@ = binds looser than ,");
 
 # terminator
 
+# superscript exponentiation
```
zoffixznet: Could you wrap this chunk in a block and give it its own `$i`. Right now it's re-using the variable used by other test 180 lines prior. It's unclear what its value really is and any changes in those 180 lines may affect the variable, not knowing it's used so far down below.

So, something like:

```perl6
{
    my $i = 2;

    # superscript exponentiation
    #?rakudo 3 skip 'superscript exponent associativity not right yet RT #130414'
    ....rest of new tests..
}
```

Also, since this test doesn't throw, we can fudge it as a `todo` rather than `skip` so that the test harness alerts us when it starts to pass:

```perl6
is(2²**3, 256, "mixed exponent does right associative");
```

zoffixznet:
Thanks!

PR #211, opened 2017-01-02
Added test for resolved bugfix RT #130474

dugword:
Tests for bug fix: https://rt.perl.org/Public/Bug/Display.html?id=130474


zoffixznet:
Thanks. New files also need to be included in [spectest.data](https://github.com/rakudo/rakudo/blob/nom/t/spectest.data). I'll add this one myself; if you have more coming, submit a PR to rakudo adding it in t/spectest.data too.

PR #212, opened 2017-01-02
Added test to fail on invalid port or family values

dugword:
For RT issues:
https://rt.perl.org/Ticket/Display.html?id=130475
https://rt.perl.org/Ticket/Display.html?id=130473


dugword:
Pull request to update spectest.data: https://github.com/rakudo/rakudo/pull/986

PR #215, opened 2017-01-04
Add command line test for UTF-8 decoding of @*ARGS

flussence:
Testcase for the bug mentioned in MoarVM/MoarVM#482.

I couldn't get the bug to reproduce with the is_run sub the rest of this file uses, so I borrowed the `is run($*EXECUTABLE, '-e'` pattern from another test file.


zoffixznet:
Thanks a lot!

PR #225, opened 2017-01-21
Add tests for RT #130554

smls:
Please review carefully, as I'm not accustomed to adding stuff to roast.

### DIFF
--- a/S03-operators/range.t
+++ b/S03-operators/range.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 125;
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks.

While it doesn't matter much in this case, `is-deeply` is a better choice than `is` for roast. For example your tests would still pass even if resulting Range was a 1-item Str rather than a list of items.

smls:
I used the same format as the existing tests right above them.  
But I will keep that in mind for future tests.

PR #227, opened 2017-01-31
Test moarvm 234

dogbert17:
Tests for fixes:
https://github.com/MoarVM/MoarVM/commit/8b195f641c
https://github.com/MoarVM/MoarVM/commit/17acfbe617

without them the test will SEGV


zoffixznet:
Thanks

PR #228, opened 2017-02-06
add fudge-usage-stat.pl to show usage of fudge implementations, verbs…

ronaldxs:
…, versions etc.

As described in my comment on Issue #226, here is my script for collecting stats on usage of fudge directives.  I think it would be better for roast to have a bin/ directory and for this, and later other, scripts to go there.  While I won't take the initiative on adding a bin/ directory I will certainly move this script there if asked.

### DIFF
--- a/fudge-usage-stat.pl
+++ b/fudge-usage-stat.pl
@@ -1,0 +1,121 @@
+#!/usr/bin/env perl
+
+######################################################################
+# fudge-useage-stat.pl
+#
+# Usage stats of fudge implementations, versions and verbs for roast.
+#
+######################################################################
+
+use strict;
+use warnings;
+
+use File::Find;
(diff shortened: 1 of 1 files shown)


fudge-usage-stat.pl
```diff
@@ -0,0 +1,126 @@
...
+use Text::Table;
+use Cwd qw(cwd);
+
+my $find_dir = @ARGV ? canonpath(shift) : '.';
+
+if ( scalar(my @a =  <$find_dir/S0[1-9]*>) < 10 ) {
+    die <<"USAGE";
+Usage: $0 [roast directory]
+
+   Intended to run against directory with synopses but not many
```
vrurg, line 24: At some point we'll drift away from binding to synopses. Just 'tests' would be good enough.

PR #229, opened 2017-02-06
Fix typo misspellings of fudge implementation niecza

ronaldxs:
As requested in comment on Issue #226.

### DIFF
--- a/S03-feeds/basic.t
+++ b/S03-feeds/basic.t
@@ -113,12 +113,12 @@ plan 23;
 }
 
 # feed and Inf
-#?nieza skip "unhandled exception
+#?niecza skip "unhandled exception
 {
   dies-ok { my @a <== 0..Inf }
 }
 
-#?nieza skip "Unhandled exception"
+#?niecza skip "Unhandled exception"
 {
   my $call-count = 0;
   my @a <== gather for 1..10 -> $i { $call-count++; take $i };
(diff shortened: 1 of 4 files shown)


niner:
I had a look at one of the misspellings to find out when it was added. That would be May 2013. Considering that niecza saw commits until January 2014, I wonder if the fudge is actually needed? Someone would have stumbled across this in those 7ish months, wouldn't they?

zoffixznet:
> I wonder if the fudge is actually needed? 

I wonder if any of them are needed. This PR is not the first time someone tried to either fix niecza fudges or were adding niecza fudges to their code. Seems it'd be beneficial to remove them all. In the unlikely even that niecza is resurrected, there's the fudge tool, isn' there?

coke:
@niner  - I think it's right to merge this, as zoffix just did - but we could definitely consider removing all the niecza fudges at this point. (I couldn't get it to work the last time I tried, so no easy way to verify which fudges were needed or not) - without an active developer now, probably better to remove them all rather than fudge them properly.

niner:
@coke yes, I'm in your camp with this. Those fudges are not contributing anything at this point. And in the unlikely event of a niecza resurrection, a removal commit can always be reverted. One would probably have to add lots of new fudges anyway and some will have become obsolete due to rewriting of the tests (e.g. during the GLR).

PR #236, opened 2017-02-17
Tests for self-referential things

AlexDaniel:
Some tests were deleted according to the rakudo pull request that
removes the error completely. Let's take a look at the deleted tests
one by one.

The first test checks that ｢is default(%h<foo>)｣ gives “Cannot use
variable $!name in declaration to initialize itself” error. This error
is completely irrelevant to what actually happens. See comments
on RT #121807.

Second test makes sure that ｢my $z = $z｣ is a compile-time error. It
could be so, but I see no basis for this. This particular case *has
never been a problem*. All rakudo versions since 2014.01 (which is the
first release of rakudo on moar) result in $z being Any, just like if
you left $z uninitialized.

The last test is the only one that tests for misuse (using
self-referential things with binding). It is based on a 5 year old
ticket that expected some completely different behavior. This code
does not cause rakudo to blow up, but it is still something we should
complain about compile time. After the removal of this error, this
code will now fail during run time because @foo will end up being
Mu. Not too bad, but could be better.

Yet nothing for problematic cases that are caused by what the error is
talking about.

Instead, I've added some tests that check that self-referential
arrays (and other self-referential things) are working correctly.

It would be nice to have some sort of design decision on what should
happen with binding. For example, we can say that getting Mu there is
OK. See RT #87034 for further discussions.

### DIFF
--- a/S02-types/WHICH.t
+++ b/S02-types/WHICH.t
@@ -350,7 +350,6 @@ my @exception = <
   X::Syntax::Variable::BadType
   X::Syntax::Variable::ConflictingTypes
   X::Syntax::Variable::IndirectDeclaration
-  X::Syntax::Variable::Initializer
   X::Syntax::Variable::Match
   X::Syntax::Variable::MissingInitializer
   X::Syntax::Variable::Numeric
--- a/S04-declarations/my.t
+++ b/S04-declarations/my.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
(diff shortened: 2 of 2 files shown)


JJ:
Can you please solve the conflict and check if this is still relevant?

AlexDaniel:
@JJ of course it is. If you follow the link to the original PR you'll see it. There's enough information on this issue.

Issue #235, opened 2017-02-16
Remove fudges for all implementations other than rakudo[.moar|.jvm]?


ronaldxs:
In a discussion on a pull request to fix niecza fudge misspellings it was suggested that we remove all niecza fudges - [PR 229](https://github.com/perl6/roast/pull/229).  The current tally of fudge implementation usage is below:

```
implementation Count by occurrence
kp6               1
mildew            6
niecza         1475
rakudo          756
rakudo.jvm      300
rakudo.moar     353
```

Besides niecza there are only 7 fudge specifications for implementations other than rakudo.  Is it agreed to remove all implementation fudges not related to rakudo?

usev6:
I'm in favor of such a consolidation.

coke:
Please do. Anyone coming back to niecza development (e.g.) at this point is going to have to start over regardless.

Please do it as one commit per implementation, so if someone wants to restore a particular one, it's as easy as possible for them.

ronaldxs:
In some cases #?niecza fudges also have #?DOES directives and in 11 of those cases niecza is the only fudge implementation so removing the #?niecza will leave a #?DOES with no implementation.  Since there are already 90+ existing cases of #?DOES with no implementation, I am only planning to remove non rakudo implementation directives and will not remove any #?DOES directives.  Please reply if you have any objection to my conclusion on #?DOES.

Since there are about 1500 non rakudo fudge implementation directives I had to write a program to search for #?DOES directives with either no implementation or niecza as the only implementation.  The program is a modified version of fudge-usage-stat.pl from PR #228 and I do not guarantee the counts are right but hope and believe so.

samcv:
I am for this too. Good plan!

PR #237, opened 2017-02-17
Remove artifact implimentations (niecza, mildew, kp6)

ronaldxs:
Trying to resolve issue #235. One commit for each implementation removal as requested.  Tested against 'make spectest' on recent Rakudo and passed.


ronaldxs:
If I do not hear back with any issues, will apply soon after today's release.

PR #240, opened 2017-02-20
fix fudgeandrun handling of return code from prove or perl6 run of test

ronaldxs:


zoffixznet:
> if $? and $? != 256; #  256 - rc of 1 from perl6 or prove

FWIW, perl6's return code indicates the number of failed tests:

```
zoffix@VirtualBox:/tmp/tmp.lGuseBLrpy$ cat foo.t 
use Test;
plan 3;
ok 1;
ok 0;
ok 0;

zoffix@VirtualBox:/tmp/tmp.lGuseBLrpy$ perl6 foo.t; echo $?
1..3
ok 1 - 
not ok 2 - 

# Failed test at foo.t line 4
not ok 3 - 

# Failed test at foo.t line 5
# Looks like you failed 2 tests of 3
2
zoffix@VirtualBox:/tmp/tmp.lGuseBLrpy$ 
```

ronaldxs:
prove and perl (5) with `use Test;` doesn't seem to work that way although with "use Test::More;" perl (5) does.  I will try to post a better PR soon.  Agreed that this should not be merged.

PR #244, opened 2017-02-23
Test to cover RT #130712

Altai-man:


zoffixznet:
`X::AdHoc` is a generic exception and in such cases we should use `Exception` in the test, in case we want to have a special exception for this case.

I'm also a beat weary of testing for textual content of errors in the Perl 6 Specification: we (or another implementation) may come up with better wording for this error and the test would be blocking such a change.

I'd recommend placing such tests into Rakudo's test suite instead, in https://github.com/rakudo/rakudo/blob/nom/t/05-messages/01-errors.t

(P.S.: I'm yet to make that file run as part of `make test`; so currently you need to use `make t/05-messages/01-errors.t` to run it).

Altai-man:
I'll use `Exception` instead of `X::AdHoc` then, thanks.

I avoided any textual context, it checks only for a presence of "returns" in the message. It is hard to imagine how you can reword "maybe you want "returns" instead" without "returns" part.

Sure, I'll move the test now.

zoffixznet:
> It is hard to imagine how you can reword "maybe you want "returns" instead" without "returns" part.

What if someone makes a much faster implementation of Perl 6, but at the cost that they can't recommend `returns` in this case?

> Should I send a PR to rakudo instead of roast then?

I would recommend that, yes.

Altai-man:
>What if someone makes a much faster implementation of Perl 6, but at the cost that they can't recommend returns in this case?

Well, we have fudging in that case... But I see your point.

I'm closing this PR then, thanks for your review!

Issue #230, opened 2017-02-07
`Iterable.flat` is undertested


zoffixznet:
`grep -FR '.flat'` and `grep -FR 'flat'` bring up a few usages of `Iterable.flat` but they seem to be more about flattening something else while testing something else and I'm failing to find any solid tests for `flat` itself.

PR #245, opened 2017-02-25
Add tests for `flat`.

hchienjo:
Add some basic tests for `flat`.

Proposed fix for #230.

### DIFF
--- a/S32-list/flat.t
+++ b/S32-list/flat.t
@@ -1,0 +1,14 @@
+use v6;
+use Test;
+
+plan 3;
+
+{
+  my int @a = 1..5;
+  is(@a, @a.flat, '.flat on an already flat list works');
+
+  my @b = ((1,2,3), (4));
+  is(@b.flat, @b, '.flat does not recurse into sublists');
+
+  is((1 .. 5).flat, @a, '.flat flattens a Range');
+}


lizmat:
I'm not sure that stringyfing results in this context is the right thing to do:

$ 6 'say ("a",<b c>).flat; say <a b c>.flat'
(a b c)
(a b c)

Tests should probably be done using is-deeply

hchienjo:
@lizmat So I should replace `is` with `is-deeply`? I borrowed inspiration from `sort.t`

lizmat:
Sorry, I guess I should have been clearer: yes, I'd prefer the test being done with is-deeply.  The tests in sort.t do not really test the result of structured data, afaik, so it's ok in those cases I think.

zoffixznet:
> So I should replace is with is-deeply

Yes, as it tests the structure and types as well, using [`eqv` operator](https://docs.perl6.org/routine/eqv), whereas `is` tests stringified versions.  For example `is (1, 3, (4, [6])), '1 3 4 6'` is a passing test, despite wildly different given and expected.

Some more tests can be added, if you're up for it. One thing's currently missing are lazy iterables:

```
    is-deeply ((3, 4, 5), lazy 5...9).flat, (3, 4, 5, 5, 6, 7, 8, 9), '.flat flattens lazy lists';
```

Would be nice to test infinite sequences too, but off-hand I'm unsure how...

Also, along with the `.flat` method, we have a `flat()` subroutine that wants to be tested as well.  If you have a lot of tests, you can [loop over both of them](https://github.com/perl6/roast/blob/f1c074e9d4957ef896bb7e48d7b0fe98a625d550/S32-str/parse-base.t#L15) and reuse the tests.

Lastly, since you're adding a new file, after this PR is merged, be sure to submit a PR to Rakudo, to add this file [to the list of files run during spectests](https://github.com/rakudo/rakudo/blob/nom/t/spectest.data).

b2gills:
You need to add a comma to create a one element list as these are both equivalent

    my @b = ((1,2,3), (4));
    my @b = ((1,2,3), 4);

It needs to be written as

    my @b = ((1,2,3), (4,));

Also `flat` does go into sublists

    my \b = ((1,2,3), (4,));
    say flat b; # (1 2 3 4)

It doesn't if those sublists are itemized

    my \b = ( $(1,2,3), $(4,) );
    say flat b; # ((1 2 3) (4))

An array variable itemizes it's values just like those shown.

zoffixznet:
@hchienjo ping. Is this PR still alive? See above comments for the proposed changes to it.

hchienjo:
Yes of course.

zoffixznet:
Thanks.

(will commit the changes proposed above manually)

PR #249, opened 2017-03-05
Tests for ≤ ≥ ≠ unicode ops

AlexDaniel:
Rakudo PR: https://github.com/rakudo/rakudo/pull/1032

### DIFF
--- a/S03-operators/misc.t
+++ b/S03-operators/misc.t
@@ -7,7 +7,7 @@ use Test;
 Tests for Synopsis 3
 =end kwid
 
-plan 44;
+plan 55;
 
 my $str1 = "foo";
 my $str2 = "bar";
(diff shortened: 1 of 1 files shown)


zoffixznet:
We should add tests that also test the chaining of the same operator. For example, without the added grammar in core, this gives wrong result, because the op isn't chained:

```
<Zoffix> m: sub infix:<≤> {$^a <= $^b }; say -5 ≤ -4 ≤ -3
<camelia> rakudo-moar e114d5: OUTPUT: «False␤»
```

It's hard to tell whether the last two tests would catch this issue.

AlexDaniel:
I've added one test for it. Not sure if all possible variants (and combinations) need to be tested.

zoffixznet:
You've added `≤`, but not the other ops you're testing 😼

AlexDaniel:
@zoffixznet IMO these tests don't add anything new to the picture, but I've added two more anyway.

For example, it is also possible to say that any test that contains ``<=``/``>=``/``!=`` should be duplicated with the unicode variant (and same thing should be done with ``o``, ``×``/``÷`` and other alternative ops). Maybe these can be autogenerated, I don't know.

zoffixznet:
> IMO these tests don't add anything new to the picture, but I've added two more anyway.

But that's only because you're letting your knowledge of the implementation seep into the tests. If another impl fails to implement other ops or if Rakudo makes a change that removes chaining of the ops you did not test, the tests would not catch that.

zoffixznet:
Thanks.

PR #253, opened 2017-03-15
Test for 130998

FCO:


samcv:
Would have liked to have some information other than just the RT number in the commits or the pull request. But thank you for adding this.

PR #255, opened 2017-03-16
#130610: Triangular reduce with &infix:<,> and a list of lists should…

FCO:
#130610: Triangular reduce with &infix:<,> and a list of lists shouldn't flatten


lizmat:
"shouldn't flatern"  ??

FCO:
Sorry, terrible english... :(

zoffixznet:
Thanks.

PR #256, opened 2017-03-18
Add tests for #130940

FCO:


S05-substitution/subst.t
```diff
@@ -625,4 +625,6 @@ is-deeply (S:g/FAIL// with 'foo'), 'foo',
 is-deeply (eager <a b c>.map: {S/a/x/}), <x b c>,
     'S/// can be used in map (does not reuse a container)';
 
+try { 42.subst-mutate: Str, Str }; pass "Cool.subst-mutate with wrong args does not hang";
+
```
zoffixznet: This test does not cover the bug, since it doesn't hang even before the fix.

Should probably be `($ = 42).subst-mutate: Str, Str`

zoffixznet:
Thanks

PR #258, opened 2017-03-21
Test triangle reduce with listinfix op. RT #131009

FCO:


zoffixznet:
These test should test the return types and structure, so you should be using `is-deeply` instead of joining the output or using `.perl`. Also, test descriptions are missing.

PR #262, opened 2017-04-08
Test fix for not dying on 'for $x()'

bduggan:
This adds a test to ensure that 'for $x()' doesn't die with an error about this being Perl 5 code.

Rt ticket: https://rt.perl.org/Public/Bug/Display.html?id=131076#ticket-history

Commit that fixes this: rakudo/rakudo@87fe800d1d

### DIFF
--- a/S10-packages/basic.t
+++ b/S10-packages/basic.t
@@ -5,7 +5,7 @@ use v6;
 use lib 't/spec/packages';
 use Test;
 
-plan 81;
+plan 82;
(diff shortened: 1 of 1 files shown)


zoffixznet:
You need to bump the `plan` on line 8

bduggan:
Thanks -- fixed and force-pushed.

zoffixznet:
Thanks.

PR #267, opened 2017-05-10
[unicode-grant] Commit New version of GraphemeBreakTest.t

samcv:
New script tests the contents of each grapheme individually from
the GraphemeClusterBreak.txt file from the Unicode 9.0 test suite.

Previously we only checked the total number of ‘.chars’ for the
string as a whole. Here we actually check the string length as well
as that each grapheme contains the exact correct codepoints
in the correct order and correct graphemes.

This new test uses a grammar to parse the file and generally is much more
robust than the previous script.

Running the parse class generates an array of arrays where the index
of the outer array indicates which

[[10084, 776], [9757]] would indicate the 0th grapheme is made up of
cp's 10084 and 776 and the 1st grapheme is made up cp 9757.


samcv:
Looks like github is trying to show a diff for a file that I deleted and put in the new one. To see the new test file go here, instead of seeing a diff against a really long file that was deleted: https://github.com/samcv/roast/blob/ce6eb28b17f3722cf93724fa768f42614d9b4d2e/S15-nfg/GraphemeBreakTest.t

PR #269, opened 2017-05-15
Add tests for Proc::Async.ready

the-eater:


S17-procasync/nonexistent.t
```diff
@@ -14,5 +14,9 @@ dies-ok { react { whenever $stderr { } } },
     'Trying to tap STDERR of an async process that does not exist signals failure';
 lives-ok { $prog.close-stdin },
     'Closing stdin of an async process that does not exist is harmless';
+dies-ok { await $prog.ready },
+    'Ready promise for a process that does not exist is broken';
 dies-ok { await $promise },
     'Promise for a process that does not exist is broken';
+
+cmp-ok $prog.ready.cause, 'eq', X::OS.new(os-error => "no such file or directory");
```
zoffixznet: > cmp-ok $prog.ready.cause, 'eq', X::OS.new(os-error => "no such file or directory");
  
Rather we don't test for exact string content in roast. I'd write that as:

`isa-ok $prog.ready.cause, X::OS`

the-eater: Cool! will fix :)

zoffixznet:
Thanks.

PR #273, opened 2017-06-01
Added test for RT 131381

danielcliffordmiller:
Wrote a test for RT 131381 that currently fails.
https://rt.perl.org/Public/Bug/Display.html?id=131381

### DIFF
--- a/S12-subset/subtypes.t
+++ b/S12-subset/subtypes.t
@@ -267,6 +267,14 @@ ok "x" !~~ NW1, 'subset declaration without where clause rejects wrong value';
     dies-ok { EVAL('f(-2)') }, 'Cannot violate Int::Positive constraint';
 }
 
+# RT #131381
+{
+    subset PInt of Int where { $_ > 0 };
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Two quick things. You need to increment the plan count (https://github.com/danielcliffordmiller/roast/blob/d4aea7748a4e62e6ed218ab7f7e9bb87f561d005/S12-subset/subtypes.t#L7) by the number of added tests . Also, if the test fails, you need to fudge it with a todo on the like before the lives-ok (e.g., `#?rakudo todo 'RT #131381'`).

zoffixznet:
Thanks. I think there must be no space in the fudge tho, so it'd be:
```perl6
#?rakudo todo 'blah blah'
```

zoffixznet:
@danielcliffordmiller BTW, add yourself to CREDITS: https://github.com/rakudo/rakudo/blob/nom/CREDITS

Currently, you'll be listed in release announcement as `danielcliffordmiller`

PR #292, opened 2017-08-02
Test for #131827

Altai-man:
Lizmat++ for the actual bugfix. Closes https://rt.perl.org/Public/Bug/Display.html?id=131827


zoffixznet:
Thanks

PR #295, opened 2017-08-18
additional failing heredoc tests

molecules:
10 of these 11 new tests show tabs getting replaced with spaces when they shouldn't be.


zoffixznet:
Comments [from IRC](https://irclog.perlgeek.de/perl6/2017-08-18#i_15039908):

```
not a fan of those super-long lines. Even on github they don't fit on screen and require scrolling :/

well, the PR looks good to me. Just merge it (after getting the RT number? and maybe after making line lengths shorter? :))
```

PR #305, opened 2017-09-04
Fix test that expected pre-GLR behavior.  RT##124572.

skids:


skids:
Alters test, so second opinion needed for merge.

Next PR will be for 6.c-errata

lizmat:
I'm pretty sure `if 0` gets compile time folded, so perhaps you should make that an `if $a` with `$a` being something falsy.  Also, perhaps test with `is-deeply ..., (42,43), ...` ?  Depending on stringification is a common source of tests passing when they shouldn't.

skids:
I could add some other tests that test a non-compile-time-folded if statement to master... probably both cases deserve testing, but keeping these particular tests as close to their original intent for 6.c-errata purposes is best.

Not sure a listy is-deeply test tells us more than checking that 43 ended up at index 1.

zoffixznet:
Perhaps some comments should be added to explain the `now == now` business and why it's done. Also, may as well subtract more than just 1 second, to completely eliminate possible races where the two values will match.

zoffixznet:
👍 Thanks!

PR #306, opened 2017-09-04
Fix test that expected pre-GLR behavior.  RT##124572.

skids:


skids:
Alters 6.c-errata, so second opinion needed for merge.
See last PR for corresponding master branch commit.

zoffixznet:
👍 on the change, as the test is wrong. `Empty` is the correct return value, not an empty `List`

Probably can be done by just cherry picking the [improved commit](346f058338718f0e6777c5c90e50597b29441b52) that went into master:

```bash
git checkout 6.c-errata
git cherry-pick 346f058338718f0e6777c5c90e50597b29441b52
git push
```

zoffixznet:
I did the cherry pick and committed the change. Thanks.

PR #310, opened 2017-09-11
Add test for stubbed class

Gnouc:
X::Package::Stubbed should not show line number

### DIFF
--- a/S12-class/stubs.t
+++ b/S12-class/stubs.t
@@ -3,7 +3,7 @@ use Test;
 
 # L<S12/Classes/You can predeclare a stub class>
 
-plan 8;
+plan 9;
 
 eval-lives-ok q[ class StubA { ... }; class StubA { method foo { } }; ],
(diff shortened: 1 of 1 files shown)


zoffixznet:
I'm going to reject this, as it doesn't make much sense to require a language implementation to never show a line number. For example, showing the location where the stub was declared could be useful. We also tend to avoid testing exact message content, as in the past such tests proved to be too restrictive for no good reason.

If you still'd like to add a test to cover your PR, you could place it into Rakudo's test suit instead. The [t/05-messages/01-errors.t](https://github.com/rakudo/rakudo/blob/nom/t/05-messages/01-errors.t) file sounds like a fine candidate to place the test into. You can run the test file with `make t/05-messages/01-errors.t` from rakudo's checkout.

jnthn:
fwiw, I think we really should show the line where the stub was declared; it'd be much kinder to the user :-)

PR #314, opened 2017-09-14
adjust tests for closure of RT#131187

skids:


skids:
The first small commit may be worthy of a cherrypick to errata, if we care about fixing tests that are skipped there.

zoffixznet:
> if we care about fixing tests that are skipped there.

We don't. And future language releases will have all the fudges stripped from them.

PR #319, opened 2017-09-16
add test for srand has default value

Gnouc:


lizmat:
I don't think the test tests whether `srand` returns a value: you're only testing that you can call `srand` without any parameters.  Or am I missing something?

Gnouc:
@lizmat with current rakudo version, calling `srand` without any parameter raise an error. If [this PR](https://github.com/rakudo/rakudo/pull/1159) were merged, then no error raising at all.

lizmat:
Ah, duh.  Should be more awake to look at PRT's  :-)

Ok, I'm not going to judge the other PR: not sure if TimToady / jnthn have any ideas on that.  I seem to recall that the absence of a default was intentional.  And clearly we shouldn't merge this until the other one is.

Gnouc:
@lizmat Me too. I just follow the `TODO` comment in rakudo code base, you can see the old comment told that using `time` as default value.

pmichaud:
Using the system clock time as a random number seed is often considered to be an unwise choice -- especially if random numbers are somehow used for cryptosecurity purposes.  (```time``` isn't random, it's highly predictable.  If you can narrow the random number seed down to a specific year, that's only 25 bits of randomness to be dealt with, which modern computers can crack pretty quickly.  Narrow it down to a month and it's only about 21 bits of randomness.)  

It also isn't a good choice if there's a possibility if multiple instances are all started within the resolution of the result (in the case of ```time```, more than one process per second), as they will likely all get the same sequence of "random" numbers.

In general, Perl 6 probably shouldn't provide defaults that are widely considered unsafe or substandard practice.

My $0.02,

Pm

lizmat:
"Perl 6 probably shouldn't provide defaults that are widely considered unsafe or substandard practice."

jonathanstowe:
"I'm wondering if this should be documented somewhere like "sources of high quality randomness are platform specific" and punted to the ecosystem.

Gnouc:
@pmichaud yes, we all know that. But `srand` or `rand` in any language I know are never used for cryptosecurity.

I made this PR just because the TODO section tells that.

niner:
On Samstag, 16. September 2017 19:03:42 CEST Cuong Manh Le wrote:
> @pmichaud yes, we all know that. But `srand` or `rand` in any language I
> know are never used for cryptosecurity.

Oh, if only that were true :) As someone who has committed this particular sin 
(excused only by my youth back then), I can assure you that it happens.

PR #324, opened 2017-09-18
Add tests for trim* subroutines on Cool instance

Gnouc:


lizmat:
You should really consider `Cool` a role.  In my opinion, it just isn't because of historical reasons.

A `Cool.new` instance is without meaning.  So the tests as such as really without meaning, even though they pass.  It probably makes more sense to test for `Int`s, since these are also `Cool` but not `Str`.  

So, `my $c = 42` instead of `my $c = Cool.new` would do the trick.

Gnouc:
@lizmat fixed it.

zoffixznet:
>  It probably makes more sense to test for Ints,

`IO::Path` is a better candidate, since you can actually test whether trimming works.

Also, were these tests to fail, the failure would be rather non descriptive. It'd be nicer to test actual result instead of relying on behaviour of other methods

```perl6
plan 1;
subtest 'trim routines on Cool' => {
    plan 6;
    my $p = ' foo '.IO;
    is-deeply trim($p),           'foo',  '&trim';
    is-deeply trim-leading($p),   'foo ', '&trim-leading';
    is-deeply trim-trailing($p), ' foo',  '&trim-trailing';

    is-deeply $p.trim,            'foo',  '.trim';
    is-deeply $p.trim-leading,    'foo ', '.trim-leading';
    is-deeply $p.trim-trailing,  ' foo',  '.trim-trailing';
}
```

Now, if some routine returns incorrect result, you'll actually see the wanted/expected values and why it failed.

zoffixznet:
👍 Thanks!

Issue #293, opened 2017-08-14
Need tests for <ww> within word


ronaldxs:
&lt;ww&gt; is in the [language/regexes doc](https://docs.perl6.org/language/regexes#Backslashed,_predefined_character_classes) but not directly tested in roast. &lt;wb&gt; seems well tested and the tests should be similar but it doesn't look simply inverting the &lt;wb&gt; tests gives a clear and good  test suite for &lt;ww&gt;.

&lt;wb&gt; tests:  https://github.com/perl6/roast/blob/master/S05-mass/rx.t#L2089-L2125

In [nqp](https://github.com/perl6/nqp/blob/817625d7b56317f6d594d87549433ca97e215f90/src/QRegex/Cursor.nqp#L982) &lt;wb&gt; and &lt;ww&gt; happen to have separate implementations.

PR #331, opened 2017-09-23
Added tests to cover <ww> - within word

mryan:
Added tests to cover the predefined subrule <ww>.  As explained
in RT #293, word boundary, <wb> is well tested but not <ww> which
has a seperate implementation in nqp.

These tests fill that void.
Resolves: RT #293

### DIFF
--- a/S05-mass/rx.t
+++ b/S05-mass/rx.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks!

mryan:
Happy to have made a contribution.
Martin

Issue #118, opened 2016-04-30
File Tests Incomplete


zoffixznet:
Looking at [S32-io/file-tests.t](https://github.com/perl6/roast/blob/master/S32-io/file-tests.t), I see many [file tests offered by Rakudo](https://github.com/rakudo/rakudo/blob/nom/src/core/IO/Path.pm#L541) are not tested for. These include `.rw`, `.rwx`, `.z` and possibly others.

PR #342, opened 2017-10-07
Expand file tests

Tyil:
This PR intends to expand the tests in `S32-io/file-tests.t`, as referenced in #118.

### DIFF
--- a/S32-io/file-tests.t
+++ b/S32-io/file-tests.t
@@ -8,7 +8,7 @@ use Test::Util;
 
 # L<S32::IO/IO::FileTests>
 
-plan 12;
+plan 13;
 
 my %tempfiles =
     existing => make-temp-file(:content("0123456789A")),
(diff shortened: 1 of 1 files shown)


S32-io/file-tests.t
```diff
@@ -3,71 +3,249 @@ use Test;
...
 
 { # write the file first
-    my $fh = open($zero-length-file, :w);
+    my $fh = open(%tempfiles<zero>, :w);
     $fh.close();
 }
 
+{ # Create a symlink to an existing file
+    symlink(%tempfiles<existing>, %tempfiles<symlink-existing>);
+}
```
zoffixznet: Instead of creating all of these manually in the current dir and then manually cleaning them up, you should add to the top of the file (after `use v6`):

```perl6
use lib <t/spec/packages/>;
use Test::Util;
```

And that will [give you `make-temp-file` and `make-temp-dir` routines](https://github.com/perl6/roast/blob/3014999fcddc4590338ffce842476da12a6817b8/packages/Test/Util.pm#L519-L534). `make-temp-file` without `:content(...)` argument creates just a temp path so you could use that for `non-existing` file too. Alternatively, you can just `chdir` to a `make-temp-dir`'ed dir and create all of these files up in there.

This way the test file will cut down on setup code all of these files aren't left hanging around in the current dir if the spectest is interrupted.

PR #344, opened 2017-10-08
Add tests for RT#125466

bitrauser:


zoffixznet:
Need to bump the `plan` on line 7 by the number of tests you're adding.

To run the test to see if it passes under the test harness, you can do `t/fudgeandrun S03-operators/bit.t`

zoffixznet:
👍 Thanks!

PR #346, opened 2017-10-08
Add tests for RT#125215 | [BUG] 10 ** -1 is a Num, but should probably be a Rat since it can in Rakudo

bitrauser:
#125215: [BUG] 10 ** -1 is a Num, but should probably be a Rat since it can in Rakudo
https://rt.perl.org/Ticket/Display.html?id=125215


zoffixznet:
👍

PR #347, opened 2017-10-13
Add test for RT#131490 | Crash in Junction:D.BUILDALL `This type (Scalar) does not support elems`

bitrauser:
#131490: Crash in Junction:D.BUILDALL `This type (Scalar) does not support elems`
https://rt.perl.org/Ticket/Display.html?id=131490


zoffixznet:
👍 Thanks!

PR #355, opened 2017-11-14
Merge redundant test of =finish pod

W4anD0eR96:


zoffixznet:
👍 Thanks! Would you also remove it from Rakudo's list of test files? https://github.com/rakudo/rakudo/blob/master/t/spectest.data#L40

PR #356, opened 2017-11-21
Allow tests to be run outside of rakudo dir

ugexe:
Absolutify lib paths and file paths using $?FILE.IO.parent(...),
since a test can know the file layout of its own repo but not the
cwd the user is invoking a test from.

This allows you to install perl6, clone the roast anywhere you
wish, and then run the spectest with:
prove -v -e 'perl6' path/to/roast/*.t

fudging still needs to be done:
perl fudgeall rakudo.moar **/*.t > test-list-spaces.txt
perl -p -e 's/\s+/\n/g' test-list-spaces.txt > test-list.txt
prove --ignore-exit -r -j4 -e 'perl6' - < test-list.txt

This doesn't handle the spectest.data list since it does not exist
in this repo. The important step here is simply getting out of
the rakudo repo, which makes parallel build system stuff possible.

### DIFF
--- a/MISC/bug-coverage-stress.t
+++ b/MISC/bug-coverage-stress.t
@@ -1,5 +1,5 @@
 use v6.c;
-use lib <t/spec/packages>;
+use lib $?FILE.IO.parent(2).add("packages");
 use Test;
 use Test::Util;
 
--- a/S01-perl-5-integration/eval_lex.t
+++ b/S01-perl-5-integration/eval_lex.t
@@ -1,5 +1,5 @@
 use v6;
-use lib <t/spec/packages/>;
+use lib $?FILE.IO.parent(2).add("packages");
 use Test;
 use Test::Util;
 plan 2;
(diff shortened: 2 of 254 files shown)


ugexe:
Waiting until after the release to merge to not slow things down.

niner:
Wow, I like this!

ugexe:
fwiw this almost works without an installed perl6 but there is a build issue. The pre-install `perl6` command/runner passes `--nqp-lib=blib` along to moar. I did a hack where I absolutify blib  [here](https://github.com/rakudo/rakudo/blob/master/tools/build/create-moar-runner.pl#L8) but adding the prefix as a Makefile var so it can be done inside the Makefile would also work. However I was only ever to get things into a state where A) perl6 worked when installed -OR- B) perl6 worked when **not** installed (but regardless of cwd)... never both.

This is assuming all use of `$*EXECUTABLE` is switched to `$*EXECUTABLE.absolute inside of spectest` (I did some already in this) so the execname passed to moar does not inadvertently cause failures for `run "./perl6", :cwd('empty-dir')`

W4anD0eR96:
I like this.

`Test::Util` is in `roast/package`. with current `use <t/spec/package>`, it's naughty to add it into lib path, when you are in dir `roast`

PR #358, opened 2017-11-25
Add test to Supply.zip

FCO:


S17-supply/zip.t
```diff
@@ -44,7 +44,24 @@ for ThreadPoolScheduler.new, CurrentThreadScheduler -> $*SCHEDULER {
...
+subtest {
+    my $*SCHEDULER = ThreadPoolScheduler.new;
+    react {
+        whenever Promise.in(1) {
+            flunk "zip timed out";
+            done;
+        }
+        whenever Supply.zip: Supply.from-list(^2), Supply.interval: .01 -> $l {
+            if $l eqv ((0, 0) | (1, 1)) {
+                pass "zip is ok"
```
jnthn: I think this could use a better description: "zip terminates when the first of its input supplies terminates".

S17-supply/zip.t
```diff
@@ -44,7 +44,24 @@ for ThreadPoolScheduler.new, CurrentThreadScheduler -> $*SCHEDULER {
 
     throws-like( { Supply.zip(42) },
       X::Supply::Combinator, combinator => 'zip' );
-
 }
 
+subtest {
+    my $*SCHEDULER = ThreadPoolScheduler.new;
+    react {
+        whenever Promise.in(1) {
```
jnthn: I'd bump this up to 5 seconds or so, just in case; systems can be under unusual load during spectest runs.

S17-supply/zip.t
FCO: do you think that for the pass message? or on subtest message?

jnthn: Oh, I just noticed there's a better one on the subtest; I missed it entirely. So, it's fine. :-)

PR #361, opened 2017-11-30
add new tests for fix to rakudo repo issue #1282

tbrowder:
accompanies rakudo PR #1287

### DIFF
--- a/S26-documentation/07c-tables.t
+++ b/S26-documentation/07c-tables.t
@@ -1,0 +1,68 @@
+use v6;
+
+use MONKEY-SEE-NO-EVAL;
+use Test;
+
+my $f = './.tmp-test-file';
+
+my $code = gen-test($f);
+
+EVAL $code;
+
+##### subroutines #####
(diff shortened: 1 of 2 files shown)


S26-documentation/07c-tables.t
```diff
@@ -0,0 +1,68 @@
...
+    $fh.print: q:to/HERE/;
+       ---+---+---
+          | X | O
+       ---+---+---
+          |   | X
+
+
+    $r = $=pod[0];
+    is $r.contents.elems, 3;
+    is $r.contents[0].join(','), 'X,O,';
```
samcv: I'm not sure why you use `.join(',')` here instead of using is-deeply and testing on an array?

tbrowder: Well, I've been using that construct for tables since I started because existing tests used it.  It seems natural because one can see easily what the table row results should be.

samcv:
Ok, merged! Thanks!

PR #362, opened 2017-12-01
Test for RT #125488

W4anD0eR96:
So far phase POST has had problem when in-loop

I'm not sure if the plan number is right(for I fudge one test), besides, if I fudge the test in the right way.

It should be mention that POST not always see no outer $_

https://github.com/perl6/roast/blob/e10d7819126d10a2731ecf00f0dcfd5abf595b56/S04-phasers/pre-post.t#L81-L95

### DIFF
--- a/S04-phasers/in-loop.t
+++ b/S04-phasers/in-loop.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 11;
+plan 14;
 
 # TODO, based on synopsis 4:
 #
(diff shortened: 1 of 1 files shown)


AlexDaniel:
POST should not see $_, at least from what I can understand. $_ in POST is set to the return value of the block:
```bash
perl6 -e 'say (for ^2 { POST .say; 42 })'
```
```
42
42
(42 42)
```

W4anD0eR96:
@AlexDaniel am I supposed to remove this subtest on POST
https://github.com/perl6/roast/blob/4355702fc69d7bfa9cda8b808c885e881f646a9d/S04-phasers/in-loop.t#L163-L169
and say that the ticket is totally resolved?

AlexDaniel:
Yeah.

PR #363, opened 2017-12-01
add tests for fix for RT #126740

tbrowder:


tbrowder:
Added tests for RT #126742 and RT #130477 (which were fixed with additions to rakudo PR #1291.

PR #368, opened 2017-12-12
`no strict` autovivification 2 level hash

ronaldxs:
RT#125183: https://rt.perl.org/Ticket/Display.html?id=125183

I think the test description could use improvement but don't know how to describe the operation better than "postcircumfix {}".


zoffixznet:
👍 Thanks!

PR #373, opened 2017-12-31
Test Issue 1204 (Strip slip in quotewords)

jstuder-gh:
This tests for bugs exhibited in
[Issue 1204](https://github.com/rakudo/rakudo/issues/1204).

Corresponds with development in Rakudo:
    [Commit ad684de383bb0c0ff051d7b0901583a9b5c2e398](https://github.com/rakudo/rakudo/commit/ad684de383bb0c0ff051d7b0901583a9b5c2e398)
    [PR 1350](https://github.com/rakudo/rakudo/pull/1350)

This does not provide cover every variation of the quoting syntax, but
hopefully provides sufficient coverage.

### DIFF
--- a/S02-literals/allomorphic.t
+++ b/S02-literals/allomorphic.t
@@ -7,7 +7,7 @@ use Test::Util;
 
 # L<S02/Allomorphic value semantics>
 
-plan 107;
+plan 110;
 
 ## Sanity tests (if your compiler fails these, there's not much hope for the
 ## rest of the test)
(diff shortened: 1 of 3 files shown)


zoffixznet:
👍 Thanks!

For record, ` is « "$x" », '5 6', ` doesn't test whether return value is a single string. It could return `(5, 6)` and the test would still pass, because `is` routine stringifies.  The `is` routine is almost always a wrong tool for the job.

The `is-deeply` routine should be used (I already changed it in 
https://github.com/perl6/roast/commit/22f00c7613a15c67845595e7d6acbceb830a68ca )

jstuder-gh:
Oh I didn't know that. I'll keep that in mind for the next time.

PR #379, opened 2018-01-23
Add a test for picking a Bag with an argument smaller than the Bag size.

lefth:
For issue #1438, which was just fixed today.


zoffixznet:
👍 Thanks!

PR #381, opened 2018-01-27
Rakudo Issue 1304: Unfudge now passing negated chain tests on JVM

jstuder-gh:
See <https://github.com/rakudo/rakudo/issues/1304>.

Depends on <https://github.com/perl6/nqp/pull/402> and <https://github.com/rakudo/rakudo/pull/1448>


zoffixznet:
👍 Thanks!

PR #382, opened 2018-01-29
Add tests for List.rotor with Whatever and Inf

taboege:
These two tests need something to the effect of https://github.com/rakudo/rakudo/pull/1454 to succeed.


lizmat:
Thanks!

PR #393, opened 2018-03-07
Skip repeated socket close test on windows rather than hang.

ronaldxs:
Skip on windows and WSL workaround for now to avoid hanging.  As mentioned in patch there is an RT and if we note workaround on ticket someone could come back and fix more fully later.


zoffixznet:
:+1: Thanks!

Issue #395, opened 2018-03-09
S16-io/eof.t broken from run-with-tty/script command on WSL


ronaldxs:
The script command is dodgy on WSL (windows subsystem for linux) and breaks testing in different ways depending on your version.  On ubuntu 16.04 the test correctly skips but before skipping it tries to run the 'script command' which messes up the terminal.   'stty sane' works for me.  This problem seems scheduled to be fixed and is documented here https://github.com/Microsoft/WSL/issues/2053 and here https://github.com/Microsoft/WSL/issues/129 .  S16-io/eof.t depends on run-with-tty from Test::Util and thus the 'script' command.  I no longer have ubuntu 14 but remember that the same test hung in that wsl version because of different problems with the script command.

I am suggesting that the test be fudged to skip for wsl and later we can look for a way to test for a wsl version/release that supports the script command properly.

zoffixznet:
:+1: on fudging for wsl

PR #397, opened 2018-03-10
S16 io eof wsl script fudge

ronaldxs:
See issue #395.  Since we change Test::Util run-with-tty we also effect S32-io/out-buffering.t as well as S16-io/eof.t.  On WSL/ubuntu 16 out-buffering.t also breaks the terminal but it ran OK on WSL ubuntu 14 since, unlike eof.t, the script process did not read from stdin.  Having the out-buffering.t test run on ubuntu 14 does not seem FTM worth a more complicated patch.

### DIFF
--- a/packages/Test/Util.pm
+++ b/packages/Test/Util.pm
@@ -343,6 +343,10 @@ sub fails-like (
 sub run-with-tty (
     $code, $desc, :$in = '', :$status = 0, :$out = '', :$err = ''
 ) is export {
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks!

If there are any more WSL detections that'd need to be made, we should think about adding some `is-OS-WSL` utility routine to Test::Util instead of having a long line with `$*DISTRO` and `$*KERNEL` each time (or, if there's enough demand, some easier way to detect WSL in Rakudo)

PR #399, opened 2018-03-11
Fix attribute mutator tests and add comment discouraging use

ronaldxs:
There seems to be more than one school of thought about using attribute mutators.  There is the @jnthn school of thought discouraging their use as mentioned in the header comment referring to the [6guts blog post](https://6guts.wordpress.com/2016/11/25/perl-6-is-biased-towards-mutators-being-really-simple-thats-a-good-thing/) and the @zoffixznet school of thought which seems more tolerant based on his [recent advent post](https://perl6advent.wordpress.com/2017/12/02/perl-6-sigils-variables-and-containers/) (the end of the section of the post on "Customizing").  I don't know which school of thought wins but believe we can do better than leaving the test broken.

### DIFF
--- a/S12-attributes/mutators.t
+++ b/S12-attributes/mutators.t
@@ -10,42 +10,64 @@ use v6;
 
 use Test;
 
-plan 10;
+plan 14;
 
-our $count = 0;
+our ($count-s, $count-m) = (0, 0);
 
 class MagicVal {
     has Int $.constant;
-    has Int $.varies = 0;
 
-    method varies is rw {
-        $count++;
+    # if you declare varies-m after varies-s you get
+    # P6opaque: no such attribute error
+    has Int $.varies-m = 0; # mutator attempting to use methods
+    method varies-m is rw {
+        $count-m++;
+
+        return Proxy.new(
+            FETCH => method ()     { $!varies-m  },
+            STORE => method ($new) { $!varies-m = $new + 1 },
+        );
+    }
+
+    has Int $.varies-s = 0; # mutator using subs
+    method varies-s is rw {
+        $count-s++;
+
         return Proxy.new(
             # note that FETCH and STORE cannot go through the accessors
             # of $.varies again, because that would lead to infinite
             # recursion. Use the actual attribute here instead
-            FETCH => sub ($ --> Int) { $!varies },
-            STORE => sub ($, Int $new --> Int) { $!varies = $new + 1 }
+            FETCH => sub ($ --> Int) { $!varies-s },
+            STORE => sub ($, Int $new --> Int) { $!varies-s = $new + 1 }
         );
     }
+
 }
 
-my $mv = MagicVal.new(:constant(6), :varies(6));
+my $mv = MagicVal.new(:constant(6), :varies-s(6), :varies-m(6));
 
-is($mv.constant, 6, "normal attribute");
 is($mv.constant, 6, "normal attribute");
 dies-ok { $mv.constant = 7 }, "can't change a non-rw attribute";
 is($mv.constant, 6, "attribute didn't change value");
+ok($count-s == 0 && $count-m == 0, "mutators not called yet");
 
-is($count, 0, "mutator not called yet");
+is($mv.varies-s, 6, "mutator called during object construction");
+is($count-s, 1, "accessor by sub was called ");
+
+$count-s = 0;
+$mv.varies-s = 13;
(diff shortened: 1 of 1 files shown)


zoffixznet:
If the test is fixed, should the fugde be removed too?

ronaldxs:
Working on removing fudge ...

ronaldxs:
Sorry.  The fudge refers to [RT #126198](https://rt.perl.org/Ticket/Display.html?id=126198) which specifies that the problem should be solvable with methods instead of subs which my fix doesn't do.  Perhaps it may be OK to have a working test with subs and have a separate fudged test for methods?  Still trying to understand why methods are important for the Proxy hooks.

zoffixznet:
That might need review by @jnthn , but I'm 95% sure expecting `$!foo` in `my method` to refer to the class that method is defined in rather than called with is wrong. 

The ticket also says that `method` keyword is practically irrelevant and it's there to get the invocant, but you also get a `self`, which again is the invocant, not the class the method was defined in

I seen it explained in channel that `Proxy.new: :STORE(my method { $!foo })` will have `$!foo` refer to the attribute in a Proxy, not in the class that that `my method` is in and that is the case for regexes too:

```
<Zoffix__> m: class Foo { has $!x; method z { "foo" ~~ /$!x/ } }.new.z
<camelia> rakudo-moar 2a3947816: OUTPUT: «===SORRY!=== Error while compiling <tmp>␤Attribute $!x not available inside of a regex, since regexes are methods on Cursor.␤Consider storing the attribute in a lexical, and using that in the regex.␤at <tmp>:1␤------>  Foo { has $!x; m…»
```

ronaldxs:
[RT #126198](https://rt.perl.org/Ticket/Display.html?id=126198) has a status of "new" and it sounds like you are saying we may want to reject the ticket and fix the test with subs.  I certainly don't understand all this well enough to try reject the ticket.

zoffixznet:
Right, that's what I'm saying. Though maybe not reject entirely, but rather clarify the general behaviour in the docs:

```
<Zoffix__> m: my method foo { say self.^name }; foo 42
<camelia> rakudo-moar 2a3947816: OUTPUT: «Int␤»
<Zoffix__> m: my method foo { say self.^name }; foo 42; foo 42e0
<camelia> rakudo-moar 2a3947816: OUTPUT: «Int␤Num␤»
<Zoffix__> m: class Foo { has $!bar = 100; method z { my method { $!bar.say } } }; class Bar { has $!bar = 42; method show ($m) { $m(self) } }; Bar.new.show: Foo.new.z
<camelia> rakudo-moar 2a3947816: OUTPUT: «42␤»
```

That all fits nicely and works out, but then this issue exists; `$!bar` needs to be in both classes for the trick to work:
```
<Zoffix__> m: class Foo { method z { my method { $!bar.say } } }; class Bar { has $!bar = 42; method show ($m) { $m(self) } }; Bar.new.show: Foo.z
<camelia> rakudo-moar 2a3947816: OUTPUT: «===SORRY!=== Error while compiling <tmp>␤Attribute $!bar not declared in class Foo␤at <tmp>:1␤------> { method z { my method { $!bar.say } } }⏏; class Bar { has $!bar = 42; method sho␤    expecting any of:␤        horizontal w…»
```

And another semi-problematic issue is using a detached method like this lets you easily poke into the guts of an object.

I've no idea what the actual behaviour should be.

ronaldxs:
I am still not clear on using methods for the Proxy hooks and [RT 126198](https://rt.perl.org/Ticket/Display.html?id=126198).  In an IRC discussion (on possibly related [RT 130403](https://rt.perl.org/Ticket/Display.html?id=130403)) it was mentioned 'the magic of Proxy is hardwired into it's new method' https://irclog.perlgeek.de/perl6/2016-12-25#i_13797157 .  The Zoffix++ Proxee module gets around all this by not suggesting methods for FETCH/STORE hooks and doing away with the extra parameter that may have made methods of interest for the hooks to begin with.  I also noticed that Proxy constructor magic may be responsible for a related symptom in [RT 124057](https://rt.perl.org/Ticket/Display.html?id=124057).

I talked to @masak about [RT 126198](https://rt.perl.org/Ticket/Display.html?id=126198) holding up testing of Proxy attribute mutators and he seemed sympathetic to this patch which keeps the fudge for the RT but tests and documents a working implementation of the feature.

zoffixznet:
👍 Thanks!

Issue #217, opened 2017-01-05
Tests needed for using method invocant to distinguish between class and object methods


ronaldxs:
```Perl6
class C {
    multi method f(::?CLASS:U:){say "class method"}
    multi method f(::?CLASS:D:){say "object method"}
}
C.f;       # says class method
C.new.f;   # says object method
```
I see some other invocant related tests but not for this use case that appears in S12, mentioned in docs, and seems well enough known.

moritz++ mentioned: a proposed addition that would let you write method f(:U:) { } for a class-only method...

A comment about the proposed addition might go in with the tests.

See also [IRC discussion](https://irclog.perlgeek.de/perl6/2017-01-05#i_13862614).

AlexDaniel:
I guess this is now resolved! Thanks!

PR #409, opened 2018-04-06
Update class-and-instance.t - tests for class and instance only methods

ronaldxs:
Tests for class and instance only methods selected by invocant concreteness (eg ::?CLASS:[U|D]:).
GH roast issue #217.


AlexDaniel:
Looks good to me. Thank you!

PR #412, opened 2018-04-07
Add subtest for lazy array cmp

jstuder-gh:
Make sure lazies behave the same way their eager counterparts do.

### DIFF
--- a/S03-operators/cmp.t
+++ b/S03-operators/cmp.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 60;
+plan 61;
 
 # cmp on scalar values
 {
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Looks good to me. Thanks!

AlexDaniel:
By the way:

> Make sure lazies behave the same as their eager counterparts.

Does it mean that there's a non-lazy set of identical tests somewhere?

jstuder-gh:
Not exactly the same. Within the same file there is a grouping of tests that test the same sort of thing [here](https://github.com/perl6/roast/blob/master/S03-operators/cmp.t#L55).

I prefer the lazy version and feel it is a bit more complete but I didn't want to modify existing, working tests.

Non-lazy version:
* Uses "is" which stringifies (although it might not be a big deal in this case of More, Less, Same).
* In the Less variant of "differing at pos 1 ignoring lengths" test, the second operand is larger at pos 1 and longer; it is More by both criteria.
* No "differ at pos 1; ignore later elems"-style tests.

Ideally, I'd like to modify those tests to be the same as the lazy counterpart but I figure it's not a good idea to change these things since the tests are essentially the language spec.

PR #413, opened 2018-04-07
Update S02-types/hash_ref.t test for RT #132238

ronaldxs:
RT #132238.
Clearer error message for unclosed <> hash quote index operator - `say $<`.

### DIFF
--- a/S02-types/hash_ref.t
+++ b/S02-types/hash_ref.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 31;
+plan 33;
 
 # basic lvalue assignment
 {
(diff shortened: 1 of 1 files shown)


zoffixznet:
Not a fan of speccing textual message content in roast. We had issues with that approach in the past. This test should probably go into Rakudo's test suite, e.g. to https://github.com/rakudo/rakudo/blob/master/t/05-messages/03-errors.t instead.

Also, the tests test for `X::Comp::AdHoc`; if we ever make those typed exceptions, the tests would start to fail. Not a problem if we're putting them into rakudo's test suite, but otherwise, that should be changed to `Exception` instead, so we could use proper type later on.

ronaldxs:
Thanks - will work on change.

ronaldxs:
Moved to t/05-messages/03-errorst.t in rakudo: https://github.com/rakudo/rakudo/pull/1706

PR #423, opened 2018-05-02
Add test for smartmatch hash with Mu values.

jsimonet:


jsimonet:
Related to https://github.com/rakudo/rakudo/issues/1772

lizmat:
Thank you!

PR #425, opened 2018-05-10
Test installation with proper Distribution usage

ugexe:
The intent of this test is installing a Distribution, and the fact
it used a class called Distribution to do it should be irrelevant.
This instead installs something that *does* Distribution.

Changed s/$path/$repo-path/ for clarity since $path is used as a
variable name elsewhere.

### DIFF
--- a/S11-repository/curli-install.t
+++ b/S11-repository/curli-install.t
@@ -1,12 +1,14 @@
 use v6;
-constant $path = $?FILE.IO.parent(2).child('packages/curi-install').absolute;
-use lib "inst#$path";
+constant $repo-path = $?FILE.IO.parent(2).child('packages/curi-install').absolute;
+use lib "inst#$repo-path";
 
 use Test;
 plan 19;
 
-rm_rf $path.IO if $path.IO.e;
-$path.IO.mkdir;
+# TODO: rename this test to `curi-install.t`
+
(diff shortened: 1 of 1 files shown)


S11-repository/curli-install.t
```diff
@@ -16,16 +18,18 @@ $path.IO.mkdir;
...
     is $*REPO.loaded.elems, 0, 'no compilation units were loaded so far';
 }
-my $dist              = Distribution.new(:name<Foo>, :api<1>, :ver(v1.2.3));
-my $non-matching-dist = Distribution.new(:name<Foo>, :api<2>, :ver(v2.3.4));
-my $foo-path = $path.IO.parent.child('Foo.pm').relative;
 
-$*REPO.install($dist,              { Foo => $foo-path });
-$*REPO.install($non-matching-dist, { Foo => $foo-path });
+my $prefix   = $repo-path.IO.parent;
+my %provides = Foo => $prefix.child('Foo.pm').relative($prefix);
```
niner: Isn't this just my %provides = Foo => 'Foo.pm';?

ugexe: It could be -- I assumed this way was used originally because it was more explicit about the relation between the path/prefix and the provided file's path.

PR #443, opened 2018-07-02
Adds tests for Distribution::Hash

### LINKED ISSUE #442: Distribution::Hash should be tested
It's [already used](https://github.com/perl6/roast/blob/b82233a6973ad18d7b74a2fdd0ae8789a1977a76/S11-repository/curli-install.t#L24), and seems to be a part of the language according to the synopsis, so it might be a good idea to include some tests here.

JJ:
And sister module, CompUnit::Repository::Distribution, which was not
tested either and seems to be part of the spec. I have chosen the
directory which fitted the best, but it might need to be moved
somewhere else.

This closes #442.

Also: Adds emacs backup files to .gitignore :see_no_evil:

### DIFF
--- a/S02-literals/allomorphic.t
+++ b/S02-literals/allomorphic.t
@@ -7,7 +7,7 @@ use Test::Util;
 
 # L<S02/Allomorphic value semantics>
 
-plan 116;
+plan 117;
 
 ## Sanity tests (if your compiler fails these, there's not much hope for the
 ## rest of the test)
@@ -448,3 +448,6 @@ subtest '.Bool on allomorphs' => {
     is-deeply .so, True,  .perl for @true;
     is-deeply .so, False, .perl for @false;
 }
+
+# GH#2010
+is <0000>.comb, "0 0 0 0", 'does Str.comb take the string logic for Bool';


S11-modules/distribution.t
```diff
@@ -0,0 +1,43 @@
...
+    is $dist.meta{$meta}, %meta{$meta}, "$meta correct for Distribution";
+}
+
+my $cur-dist = CompUnit::Repository::Distribution.new( $dist);
+isa-ok $cur-dist, CompUnit::Repository::Distribution, "Object created";
+
+for <api provides name ver> -> $meta {
+    is $cur-dist.meta{$meta}, %meta{$meta}, "$meta correct for CUR";
+}
+is $cur-dist.Str, "{%meta<name>}:ver<{%meta<ver>}>:auth<{%meta<auth>}>:api<{%meta<api>}>", "Str for CURI correct";
```
ugexe: Implementation detail

S11-modules/distribution.t
```diff
@@ -0,0 +1,43 @@
...
+
+ok $cur-dist.id, "CUR Distribution has ID {$cur-dist.id}";
+
+
+rm_rf $repo-path.IO;
+
+sub rm_rf(*@files) {
```
ugexe: not used

S11-modules/distribution.t
```diff
@@ -0,0 +1,43 @@
+use v6;
+constant $repo-path = $?FILE.IO.parent(2).child('packages/distribution').absolute;
+use lib "inst#$repo-path";
+
+use Test;
+plan 12;
+
+rm_rf $repo-path.IO if $repo-path.IO.e;
```
ugexe: not needed

S11-modules/distribution.t
```diff
@@ -0,0 +1,43 @@
...
+my $prefix   = $repo-path.IO.parent;
+my %meta     = ( :name<Foo>, :api<1>, :ver(v1.2.3), :auth<perl6>, provides => { Foo => 'Foo.pm'} );
+my $dist     = Distribution::Hash.new( %meta, :$prefix);
+
+isa-ok $dist, Distribution::Hash, "Object created";
+for <api provides name ver> -> $meta {
+    is $dist.meta{$meta}, %meta{$meta}, "$meta correct for Distribution";
+}
+
+my $cur-dist = CompUnit::Repository::Distribution.new( $dist);
```
ugexe: This should not be tested in roast.

ugexe:
Distribution implementations are not ready to have their own roast tests in roast. For the most part they are already tested their primary consumer -- https://github.com/perl6/roast/blob/master/S11-repository/curli-install.t ( which this test is mostly copied from ). As such most things required by a Distribution are part of the curli-install.t test already.

S11-modules/distribution.t
```diff
@@ -0,0 +1,43 @@
...
+
+for <api provides name ver> -> $meta {
+    is $cur-dist.meta{$meta}, %meta{$meta}, "$meta correct for CUR";
+}
+is $cur-dist.Str, "{%meta<name>}:ver<{%meta<ver>}>:auth<{%meta<auth>}>:api<{%meta<api>}>", "Str for CURI correct";
+
+ok $cur-dist.id, "CUR Distribution has ID {$cur-dist.id}";
+
+
+rm_rf $repo-path.IO;
```
ugexe: not used / needed

S11-modules/distribution.t
```diff
@@ -0,0 +1,43 @@
+use v6;
+constant $repo-path = $?FILE.IO.parent(2).child('packages/distribution').absolute;
```
ugexe: this isn't even an existing path -- nor should any additional libs be needed for such a test

JJ: OK

S11-modules/distribution.t
JJ: OK

S11-modules/distribution.t
JJ: OK

S11-modules/distribution.t
JJ: OK

S11-modules/distribution.t
JJ: OK

JJ:
Seeing it now, wouldn't it make more sense if I just added a few lines to curli-install.t instead of creating a new file?

ugexe:
Yes. That is what I suggested here as well: https://github.com/perl6/roast/issues/442#issuecomment-401810363

PR #456, opened 2018-07-31
Add tests for #2143, Distribution::Resource.^mro

ribbon-otter:
This pull request adds tests which I believe cover Rakudo's [#2143](https://github.com/rakudo/rakudo/issues/2143), which the possible exception of `Distribution::Resource.Str`.

The added tests cover `Distribution::Resource.^mro`, `Distribution::Resource.perl`, `Distribution::Resource.gist`, and `.WHAT` on an instance object. However, that instance is created with `Distribution::Resource.new` (which I copied the form from [here](https://github.com/rakudo/rakudo/blob/d77a51de230516f8cb0139df08c93590e1fb5356/src/core/Distribution.pm6#L274)) instead from `%?RESOURCES`, and thus might not being completely natural. This test does not check `Distribution::Resource.Str` because I do not know if that should work or not.

I am quite willing to move my tests to a better file, or add `Distribution::Resource.Str` to the checked tests if someone who knows more can give input.

### DIFF
--- a/MISC/bug-coverage.t
+++ b/MISC/bug-coverage.t
@@ -6,7 +6,22 @@ use Test::Util;
 # This file is for random bugs that don't really fit well in other places.
 # Feel free to move the tests to more appropriate places.
 
-plan 11;
+plan 12;
+
+#https://github.com/rakudo/rakudo/issues/2143
+subtest 'Distribution::Resources works with .gist/.perl/.^mro/.WHAT' => {
+	plan 4;
+	Distribution::Resource.gist; pass '.gist';
+	Distribution::Resource.perl; pass '.perl';
+	Distribution::Resource.^mro; pass '.^mro';
+	Distribution::Resource.new(:repo("repo"), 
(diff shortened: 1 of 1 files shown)


zoffixznet:
@niner is `Distribution::Resources` OK to add to roast? I notice there aren't any other spectests for it.

ufobat:
ping?

ugexe:
`Distribution::Resources` is full of implementation details like `$.dist-id`. And frankly it doesn't need to exist since using `Distribution` to back `%?RESOURCES` can provide the same functionality while letting the `CompUnit::Repository` that supplies the `Distribution` handle any possible implementation details ( such as passing a `$.dist-id` to the Distribution for lazy loading ).

Test against `%?RESOURCES`, not `Distribution::Resources`.

AlexDaniel:
Thanks! Similar tests were added to rakudo repo (not roast) in https://github.com/rakudo/rakudo/pull/2585!

PR #461, opened 2018-08-07
Add new test for issue #2158.

xelak6:


zoffixznet:
Thanks. FWIW, R#2158 now segfaults for me instead of failing like before (but the test doesn't trigger that)

PR #462, opened 2018-08-08
Adds test for translate-nl

JJ:
And some reasonable additional tests.


JJ:
@Kaiepi well, it was not tested, so while I was on the file I decided to include that too.

zoffixznet:
These tests are broken. You're testing the literal `\` and `n` instead of the actual newline characters and the non-Windows branch is missing `:translate-nl` argument.

Also, please use `is-deeply` routine instead of `ok`, so failures would actual show the differences between  expected and got versions.

zoffixznet:
:+1: Thanks!

JJ:
Thanks for reviewing and spotting the errors :-)

PR #467, opened 2018-08-30
Test STOREing / splicing lazy Seq into native array throws

jstuder-gh:
splicing as well.

Corresponds to the following
[Rakudo development](https://github.com/rakudo/rakudo/pull/2240)

I'm not sure this the most appropriate place for the splicing tests as the
other splice-related tests are in the type-specific native array files. But I
also wanted to minimize duplication as the code can be shared between
STORE and splice tests.

If the splice tests are better elsewhere let me know and I'll move them.

### DIFF
--- a/S09-typed-arrays/native.t
+++ b/S09-typed-arrays/native.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 8;
+plan 9;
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
Looks good. You should have sufficient permissions to merge this PR yourself.

PR #483, opened 2018-10-08
Shaped array test improvements

SyrupThinker:
This pull request enables a previously disabled shaped array test set
and changes one of its tests to the behaviour discussed [here](https://github.com/rakudo/rakudo/issues/2351).

It also adds a subtest that generates tests for arrays up to n dimensions to catch bugs like these:
https://github.com/rakudo/rakudo/pull/2350
https://github.com/rakudo/rakudo/issues/2352
because they are less commonly used.

It is currently set to three because of the latter.

### DIFF
--- a/S02-types/multi_dimensional_array.t
+++ b/S02-types/multi_dimensional_array.t
@@ -13,7 +13,6 @@ plan 58;
 # L<S09/Multidimensional arrays/Perl 6 arrays are not restricted to being one-dimensional>
 
 # real multi-dimensional arrays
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks!

AlexDaniel:
FWIW these tests fail like this before the fix: https://gist.github.com/Whateverable/14316b099399e0ad24bce845621bcc9c
Not exactly the best failure mode (I was expecting just the test to fail, not the whole test file), but at least we're now testing this stuff.

PR #496, opened 2018-11-05
Tests listening on port 0 returns usable tap with whenever

jmaslak:
This tests for regressions of:
  https://github.com/rakudo/rakudo/issues/2411

### DIFF
--- a/S32-io/IO-Socket-Async.t
+++ b/S32-io/IO-Socket-Async.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 38;
+plan 40;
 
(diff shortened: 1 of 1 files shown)


Kaiepi:
Since you're looking for more elegant ways to write the test, I'd write it like this:
```perl6
my $s = IO::Socket::Async.listen('localhost', 0).tap(-> $c { });
lives-ok { await $s.socket-port }, 'can await on socket port';
cmp-ok $s.socket-port.result, '>', 0, 'socket port looks reasonable';
$s.close;
```

I feel like this is a little more legible

PR #500, opened 2018-11-27
Adding test to dynamic shaped array and dynamic object hash

FCO:


lizmat:
Thanks for spotting and adding tests!

PR #516, opened 2019-03-04
Tests for rakudo/rakudo#2169

vrurg:


vrurg:
Something went wrong on merging roast from master. Will re-open as a new PR.

PR #519, opened 2019-03-12
Fix macos watch-path

clarkema:
This PR addresses a few issues with the watch-path tests on macOS.  Unfortunately it makes the test take a little over 2 minutes to run [and it does still occasionally generate incorrect failures] but this is the only combo that regularly passes on my machine.

In the long term it would be better to find a way of waiting up to a specified timeout for each expected event, rather than just counting hits and having to find a timeout that's "mostly ok" without spending too much time waiting.

I'm uncertain about the specified behaviour in https://github.com/perl6/roast/commit/dcf672cfb2d3cb4fd81f3c028784fd24f82bb2b4.  It seems that we don't get an notification for opening an existing file for append.  That seems sensible to me, but the wording on the test doesn't make it clear whether or not that's really the expected bevahiour.

### DIFF
--- a/S17-supply/watch-path.t
+++ b/S17-supply/watch-path.t
@@ -158,7 +158,7 @@ sub macosx (:$io-path) {
         ok $handle.close, 'did the file close ok';
 
         sleep $forawhile;
-        is +@seen, 1, 'did we get an event for closing the file';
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks!

PR #523, opened 2019-03-24
Changes to conform with fix of use of Real type object in operators

vrurg:
PR: rakudo/rakudo#2790
Ticket: rakudo/rakudo#2786


lizmat:
Thanks!

PR #527, opened 2019-04-09
add tests of rounding methods (adverbs for basic rounding methods) se…

tony-o:
see:rakudo/rakudo#2826

This commit adds tests for basic rounding methods: `HALF-[UP|DOWN|EVEN|ODD] | [TO|FROM]-ZERO`.


tony-o:
Per @AlexDaniel - review is there so this request isn't merged without further discussion see [here](https://colabti.org/irclogger/irclogger_log/perl6?date=2019-04-09#l583)

PR #541, opened 2019-05-18
Add test for GH # 2480

Altai-man:
Tries to address https://github.com/rakudo/rakudo/issues/2480

I am not sure if this is all that is needed, can someone suggest something better?


SqrtNegInf:
Looks good to me.  Corresponds to the one 'live code' instance that prompted the ticket.

PR #545, opened 2019-05-31
Changed release guide to reflect the latest state of things.

vrurg:
This version is most like incomplete and most definitely contains
errors. Review is required.

### DIFF
--- a/docs/announce/6.e.md
+++ b/docs/announce/6.e.md
@@ -1,0 +1,70 @@
+
+# Announce: Perl 6.e Language Specification Release
+
+_This to be filled up upon actual release_
+
+## Authors
+
+The following people contributed to this version of the language,
+including making documentation updates and contributing work to known
+open‑source compiler implementations. The list is ordered alphabetically.
+
+If you believe your name has been erroneously omited, please contact us
+(https://perl6.org/irc), and we'll update the primary copy of this list.
+
+_This to be filled up upon actual release_
+
+## Changelog
+
+This document lists changes in Perl 6.e language
+from Perl 6.d version. A particular implementation of the language
+may contain additional changes; please consult with the changelog for your
+implementation.
+
+At the same time, a particular implementation may have had certain features
+already implemented during the previous versions period. This ChangeLog concerns
+itself with new features added to the specification on a language level and not
+the status of their implementation in a particular compiler.
+
+### Scope / Target Audience
+
+This ChangeLog is targeted towards language users, to help with preparation to use
+compilers supporting latest language version. Thus, it does not contain every minute
+change to the specification that occurred. Implementations wishing to ensure full
+compliance with the new version of the language specification should execute the
+test suite available at https://github.com/perl6/roast/ and examine any failing tests.
+
+There are new features that did not exist in the previous releases. For full details about them,
+please consult with the language documentation on https://docs.perl6.org/
+Items in *Version-Controlled Changes* section are protected by version pragma and older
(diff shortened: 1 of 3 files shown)


docs/release-guide.md
```diff
@@ -12,20 +12,59 @@ please list the main repository for those in this section:
...
 
-## Preparation
+## Terminology
+
+The following terms are used in this document and in other related papers:
+
+- _language revision_ or _revision_
+
+  A set of specifications defining a Perl 6 language variant. Revisions are represented by symbolic names consisting of
+  a single letter of Latin alphabet and may have an alias. For example, revision `c` is `Christmas`; `d` is `Divali`.
```
AlexDaniel: Maybe we should drop this? So that `c` is just `c`. I remember a lot of bikeshedding and it'd be nice to avoid it.

docs/release-guide.md
```diff
@@ -12,20 +12,59 @@ please list the main repository for those in this section:
...
+## Release Status
+
+_This section contains excerpts from [this discussion](https://github.com/perl6/problem-solving/issues/31)._
+
+Language release process is based on the fact that changes are only allowed to a revision in _review_ status. Once a
+revision is released it becomes _immutable_. From this moment on the only changes allowed are fixes. From the point of
+view of a versioning system like git would look like the following diagram:
+
+```
+             6.d tag
```
AlexDaniel: I dislike these tags because they are historical nice-to-haves but functionally there's no reason to have them. For example, a user or a dev might accidentally check out 6.d tag instead of 6.d-errata. Maybe that's not a big deal but I'd like to see a justification for having these tags.

docs/release-guide.md
```diff
@@ -91,3 +130,23 @@ ensure it works on new language version. Often, only a small part of the
 new language version is placed behind a version pragma, with the rest of the
 changes being simply clarifications to the past spec. The upgrade guide should
 focus on the version-pragma-protected changes.
+
+## Release Steps
+
+Those steps are to be done when a new language revision is stabilized and time comes for a release (_remember that `.d`
```
AlexDaniel: Maybe use something unrealistic like `x`.

vrurg: `.<rev>`? I usually use this form.

vrurg:
Any additions perhaps? I feel like its incomplete, especially the `Release Steps` section.

docs/announce/6.e.md
```diff
@@ -0,0 +1,70 @@
...
+
+#### New Parameters and Arguments
+
+#### New Routines and Operators
+
+#### New Types
+
+#### New Variables
+
+#### Clarifications of Edge Case/Coercion Behaviour
```
JJ: Would this include disambiguation of behavior? If not, maybe a new subsection would be in order.

docs/release-guide.md
JJ: c, indeed, *was* Christmas, and d *Diwali* (with w). It says "may" have an alias. Don't see any major problem with this.

docs/announce/6.e.md
vrurg: What to include into the announcement would be absolutely up to the release manager. I created a template based on `6.d` announce. The purpose of the template is to fill it in as soon as there is something to be noted in it.

docs/release-guide.md
vrurg: @JJ I have removed it anyway. Basically, the document is purely technical and it adds nothing useful to have those aliases in there.

vrurg:
Just for reference: related to perl6/problem-solving#31

PR #556, opened 2019-07-09
Remove reference to placeholder tickets from RT

usev6:
This happens in preparation of closing a bunch tickets, that have
been created back in 2015, but don't provide useful information.

Cmp. https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2019-07-09#l83


AlexDaniel:
Hey @usev6, what can be our long-term plan for todo-ed and skip-ed tests?

For example, what if we make the next [SQUASHathon](https://github.com/rakudo/rakudo/wiki/Monthly-Bug-Squash-Day) to be about roast todo-ed tests? We can then populate a separate issue tracker with issues, one per test. We make a guide on what to do with them (e.g. if the test doesn't make sense anymore, then delete it; if there are a bunch of todo-ed tests for a feature, file a single ticket and reference it from the tests; etc.). I believe the contributors will be able to do a lot of great work. It also sounds kinda fun and relatively easy.

Similar squashathons:
* 2017-12-02 MOLD Removal SQUASHathon ([results](https://gist.github.com/Whateverable/d9a9f76e6470f80be8b03a463464c6cb))
* 2018-08-04 Ecosystem unbitrot ([results](https://gist.github.com/c55010babde46922f0587d0009b632c9))
* 2019-07-06 Math sequences squashathon ([results](https://github.com/ajs/perl6-Math-Sequences/issues/30#issuecomment-509005868))

usev6:
Hi @AlexDaniel,

> what can be our long-term plan for todo-ed and skip-ed tests?

actually, I'm not sure about that. I think we have at least three different kinds of todo-ed and skip-ed tests:

1. tests for obvious bugs (those are mostly covered with existing tickets on RT or GH)
2. tests that expect a different result than the current implementation gives, but where it's not obvious if the tests are right
3. tests for not implemented features

IMHO there are a lot of tests where it's not easy (for many contributors) to see to which category they belong. Your idea to use a squashathon to sort out todo-ed and skip-ed roast tests sounds attractive, but I fear that (most of) this work needs to be done by devs from the "inner circle". (Since roast is the specification, evaluating tests sounds a lot like language design.) So, from my perspective there aren't (m)any low hanging fruits in this area. (Of course, I might be wrong.)

AlexDaniel:
I see. Thanks

usev6:
If no one objects, I'll merge this PR and close the mentioned placeholder tickets soon-ish.

PR #562, opened 2019-07-20
Test block interpolation in routine with signature-declared return value

chloekek:
https://github.com/rakudo/rakudo/issues/3070

### DIFF
--- a/S02-literals/string-interpolation.t
+++ b/S02-literals/string-interpolation.t
@@ -3,7 +3,7 @@ use Test;
 use lib $?FILE.IO.parent(2).add("packages/Test-Helpers");
 use Test::Util;
 
-plan 40;
+plan 42;
 
 # L<S02/Closures/"A bare closure also interpolates in double-quotish context.">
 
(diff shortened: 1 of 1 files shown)


S02-literals/string-interpolation.t
```diff
@@ -141,4 +141,13 @@ line 4
 
 }
 
+# https://github.com/rakudo/rakudo/issues/3070
+#?rakudo skip 'GH#3070 not yet fixed.'
+{
+    sub f(--> Nil) { say "{1}"; }
+    sub g(--> 2)   { say "{1}"; }
+    is f, '1', 'block interpolation works with signature-declared Nil return value';
```
JJ: Not sure about this. `say` returns `True`, so f should actually return True, except it will return Nil because it's defined that way. Maybe you want to check for output of f, if that's what the bug is about.

chloekek: Oh right! Thanks, the routine shouldn't call say but rather return the string.

chloekek: Fixed and confirmed that "{1}" does indeed evaluate to the string "2".

chloekek:
The Rakudo fix has been merged, should I remove the #?rakudo fudge line?

jnthn:
> The Rakudo fix has been merged, should I remove the #?rakudo fudge line?

Yes, please.

jnthn:
Thanks!

PR #566, opened 2019-08-15
Preparing for possible change of language name to Raku

vrurg:
Following perl6/problem-solving#81 discussion.


vrurg:
Also in anticipation of possible outcome of perl6/problem-solving#88 added a bit on versioning of `-errata` branches. Whichever way the discussion will go, `-errata` will most likely have versions.

PR #569, opened 2019-08-23
Add test for GB18030 encoding

ZhongnianTao:


lizmat:
Thank you!

vrurg:
This merge was premature as it depends on rakudo/rakudo#3135. I was about to merge the rakudo PR but when tested it locally it didn't fix this test which was still failing with _unknown encoding_ error. I'm giving it currently another try, but if it fails I'd propose reverting.

lizmat:
I've removed the test file from spectest.data.  Please revert *that* when the rest has been merged.

vrurg:
Thank you! My second try on the PR went ok. I'll merge finalize everything today.

PR #576, opened 2019-09-10
Add test for file change events.

vrurg:
Followup to rakudo/rakudo#3100

### DIFF
--- a/S16-io/watch.t
+++ b/S16-io/watch.t
@@ -1,0 +1,48 @@
+use v6;
+
+use Test;
+
+plan 1;
+
+# GH rakudo/rakudo#3100
+{
+    # Test if file change events are not lost.
+    my $watch-file = $*SPEC.catfile( $*TMPDIR, "watchme" );
+    my $fh = $watch-file.IO.open: :w, :0out-buffer;
+    my $done-writing = Promise.new;
+    my $done-vow = $done-writing.vow;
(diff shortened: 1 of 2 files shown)


S16-io/watch.t
```diff
@@ -0,0 +1,48 @@
...
+
+# GH rakudo/rakudo#3100
+{
+    # Test if file change events are not lost.
+    my $watch-file = $*SPEC.catfile( $*TMPDIR, "watchme" );
+    my $fh = $watch-file.IO.open: :w, :0out-buffer;
+    my $done-writing = Promise.new;
+    my $done-vow = $done-writing.vow;
+    start {
+        sleep .1; # Be overcautious.
```
jnthn: Better would be the make a `Promise` outside of the `start` block, `await` it here (with a timeout, perhaps) and then keep it after the `whenever`s in the `react` block. Then you know for sure that the subscriptions are set up.

vrurg: Last year I perhaps caught a bug in react implementation causing some strange behavior inside it. Since then I was avoiding anything but `whenever` inside it. My initial sync was with a `Promise` being kept right before `react` but that surely didn't make sense.

Thanks @jnthn !

vrurg:
I think this is to be merged when the related ticket is resolved.

PR #585, opened 2019-10-18
Test Parameter subclassing

Altai-man:
Spec test for https://github.com/rakudo/rakudo/issues/2829 as suggested by @jnthn.

Closes https://github.com/rakudo/rakudo/issues/2829


AlexDaniel:
Yup, looks good! Thank you!

PR #586, opened 2019-10-18
Test runtime-created classes being attributes

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/2607

The question at https://github.com/rakudo/rakudo/issues/2607#issuecomment-455559432 asks if the test should live in roast or rakudo suite.

My guess is that roast is the correct place, because we likely want all proper Raku compilers to be able to do what's in the test, because MOP totally allows that, so this is not a rakudo-specific issue.

### DIFF
--- a/S12-meta/classhow.t
+++ b/S12-meta/classhow.t
@@ -2,7 +2,7 @@ use Test;
 
 use lib $?FILE.IO.parent(2).add("packages/S12-meta/lib");
 
-plan 5;
+plan 6;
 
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Yeah, even though to me it does feel like something that should maybe not be in roast, we already have tests for similar functionality, so no reason not to add more. If we later decide to make these tests not part of roast we can always move them out. Interesting code and nice test, thanks!

PR #587, opened 2019-10-18
Test hyperop results in a slip

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/2482


AlexDaniel:
It makes sense, there's no real problem with this test. However, shouldn't we also test the result too? What if it lives but gives wrong output?

Altai-man:
Added value tests, is this `is` usage ok? Because `is-deeply` does not work with resulting `Seq`.

PR #588, opened 2019-10-19
Test fake-infix adverbs on a construct

Altai-man:
Just code provided by the ticket, closes
https://github.com/rakudo/rakudo/issues/1500

### DIFF
--- a/S02-types/pair.t
+++ b/S02-types/pair.t
@@ -2,7 +2,7 @@ use v6;
 
 use Test;
 
-plan 4 * 19 + 104;
+plan 4 * 19 + 105;
 
 # L<S02/Mutable types/A single key-to-value association>
 # basic Pair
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Totally fine test, however:
> \<AlexDaniel\> c: `2015.12,HEAD` `use Test; my Pair $p; is-deeply ($p .= new :key<foo> :value<bar>), :foo<bar>.Pair, ‘foo’`
> \<committable6\> AlexDaniel, `¦2015.12: «===SORRY!===␤Unknown QAST node type NQPMu␤ «exit code = 1»» ¦HEAD(e45bb34): «ok 1 - foo␤»`

So on rakudo versions with the bug the test doesn't just fail, the whole file doesn't compile. I'm not sure what's the policy on tests like this, I'm OK with it being the way it is.

Altai-man:
>So on rakudo versions with the bug the test doesn't just fail, the whole file doesn't compile. I'm not sure what's the policy on tests like this

I think if one would break a parser and the file won't compile, instead of saying "Tests are not ok", this failure is just as obvious as nok messages. And it can happen for any file, we have "This thing parses ok" tests along with semantic ones.

AlexDaniel:
Yeah, alright!

PR #591, opened 2019-10-23
Add test for user-formatted pod leading declarator blocks

tbrowder:


vrurg:
Correct me if I'm wrong, but there was a PR merged recently which is related to this. Wasn't it?

tbrowder:
Yes, Rakudo PR #3260.

tbrowder:
I'll resubmit after I can test it more.

PR #596, opened 2019-11-12
Add test to check UInt on non-numeric Str returns a Failure

mienaikage:
Via https://github.com/rakudo/rakudo/pull/3291


lizmat:
Thanks!

PR #597, opened 2019-11-15
Add a test for rakudo/rakudo#3299 Proc::Async memory corruption bug.

SirBogman:
There are 1200 iterations of a loop to try to reproduce the memory corruption bug.
Without the fix, typically fails after 500-800 iterations.
With the fix, takes about 13 seconds on my machine for all 1200 iterations.

### DIFF
--- a/S17-procasync/stress.t
+++ b/S17-procasync/stress.t
@@ -3,7 +3,7 @@ use Test;
 use lib $?FILE.IO.parent(2).add("packages/Test-Helpers");
 use Test::Util;
 
-plan 23;
+plan 24;
 
 # RT #125515
 {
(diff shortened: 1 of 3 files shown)


lizmat:
Thanks for the test!  Fails for me on MacOS.  Guess we need to wait until MoarVM is bumped?

S17-procasync/stress.t
```diff
@@ -93,3 +93,21 @@ if run :!out, :!err, «perl -e 'print 42'» {
 else {
     skip 'need `perl` to run this test';
 }
+
+{
+  my $prog   = $*DISTRO.is-win ?? 'cmd'   !! 'cat';
+  my @target = $*DISTRO.is-win ?? «/c ""» !! '/dev/null';
+
+  for ^1200 {
+    print "\r$_ ";
```
vrurg: Tests should output nothing but TAP. This `print` must no be in production.

S17-procasync/stress.t
```diff
@@ -93,3 +93,21 @@ if run :!out, :!err, «perl -e 'print 42'» {
...
+      my $proc = Proc::Async.new($prog, |@target);
+      react {
+          whenever $proc.start { done }
+          whenever signal(SIGTERM) {}
+          whenever Promise.in(5) {}
+      }
+  }
+
+  say '';
```
vrurg: Same to the above comment.

S17-procasync/stress.t
SirBogman: fixed

SirBogman:
@vrurg I switched it to using `is_run`, which I like much better. Thanks. I also updated two other tests which used the same format.

vrurg:
@SirBogman thank you! Will be merged when the next NQP bump happens.

vrurg:
Thanks, @SirBogman!

PR #598, opened 2019-11-24
Added test for P5 regex conditional expressions.

SirBogman:
Some of these tests don't work in Rakudo.

I was trying to use this for code golf.
I have no idea how to fix this.


vrurg:
@SirBogman Thank you for submitting this PR!

Could you, please, open an issue on this in https://github.com/rakudo/rakudo/issues? This would raise a flag for the need to possibly fix the issue.

PR #601, opened 2019-12-11
Add regression test for rakudo/issues/3346

dumarchie:
Add a regression test for rakudo/rakudo/issues/3346

### DIFF
--- a/S03-operators/eqv.t
+++ b/S03-operators/eqv.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 58;
+plan 59;
 
 # L<S03/Comparison semantics/Binary eqv tests equality much like === does>
 # L<S32::Basics/Any/"=item eqv">
@@ -222,6 +222,18 @@ subtest 'Seq eqv Seq' => {
         'eqv between Seqs with different end values';
 }
 
+# rakudo/issues/3346
+{
+    my $test = start {
+        my $a = (0,1).hyper;
+        my $b = (0,2).hyper;
+        die if $a eqv $b;
+    };
+
+    await Promise.anyof($test, Promise.in(5));
+    is $test.status, Kept, 'HyperSeq eqv HyperSeq';
+}
+
 subtest 'Throws/lives in lazy cases' => {
     plan 8;
 
@@ -243,19 +255,6 @@ subtest 'Throws/lives in lazy cases' => {
     lives-ok    { (1…∞)       eqv (1…3)       }, 'Seqs, only one lazy';
     lives-ok    { (1…∞).List  eqv (1…3).List  }, 'Lists, only one lazy';
     lives-ok    { (1…∞).Array eqv (1…3).Array }, 'Arrays, only one lazy';
-}
-
-# rakudo/issues/3346
-{
-    my class Null does Iterable does Iterator {
-        method iterator  { self }
-        method pull-one  { IterationEnd }
-    }
-
-    my $p = start { Null.new eqv Null.new };
-    await Promise.anyof($p, Promise.in(5));
-    ok $p.status == Kept,
-     'eqv comparison between non-Positional Iterables does not hang';
 }
 
 # vim: ft=perl6
--- a/S03-operators/eqv.t
+++ b/S03-operators/eqv.t
@@ -246,16 +246,40 @@ subtest 'Throws/lives in lazy cases' => {
 }
 
 # rakudo/issues/3346
-{
-    my class Null does Iterable does Iterator {
-        method iterator  { self }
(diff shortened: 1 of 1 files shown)


Altai-man:
The test is a bit too specific, as it relies on knowing internals (AT-POS method).
Maybe something like:
```perl6
subtest 'Iterable, not Positional' => {
    my class Null does Iterable does Iterator {
        method iterator  { self }
        method pull-one  { IterationEnd }
    }

    lives-ok { Null.new eqv Null.new },
        'eqv usage on Iterables that do not implement Positional does not hang';
}
```
would be more generic?

dumarchie:
I don't see what's wrong with using `AT-POS` as its part of the [`Positional`](https://docs.raku.org/type/Positional) API. Furthermore, without dying on positional access the test will hang given the current implementation of `multi sub infix:<eqv>(Iterable:D \a, Iterable:D \b)`. I don't think that's what you want.

I do like your test description, though.

Altai-man:
>I don't see what's wrong with using AT-POS as its part of the Positional API

I mean, it assumes that `eqv` will call `AT-POS`, but what if one day the implementation will be changed? The test should check more narrow conditions test case, I think, it being "When something implements Iterable, but not Positional, using `eqv` won't hang".

>I don't think that's what you want.

You can check such relatively simple hangs using Promises. For example, this way:

```perl6
subtest 'Iterable, not Positional' => {
    my class Null does Iterable does Iterator {
        method iterator  { self }
        method pull-one  { IterationEnd }
    }

    my $p = start { Null.new eqv Null.new };
    await Promise.anyof($p, Promise.in(5));
    ok $p.status ~~ Kept, 'eqv usage on Iterables that do not implement Positional does not hang';
}
```
This both does not assume internals of `eqv` implementation and does not hang.

dumarchie:
Ah, of course, such (ab)use of `Promise` should be fine in short-lived scripts. Thanks for sharing!

S03-operators/eqv.t
```diff
@@ -205,4 +205,17 @@ subtest 'Throws/lives in lazy cases' => {
...
 
+# rakudo/issues/3346
+{
+    my class Null does Iterable does Iterator {
+        method iterator  { self }
+        method pull-one  { IterationEnd }
+    }
+
+    my $p = start { Null.new eqv Null.new };
+    await Promise.anyof($p, Promise.in(1));
```
vrurg: Please, make it 5 as was proposed. Sometimes on heavily loaded slow setups this might result in false negative. Bigger value is fine for a failing test. While passing test won't bother at all.

dumarchie: As you wish.

niner: In that regard, why not going for 60 instead for the same reasons? 5 seconds may seem long, but once you turn on MoarVM's GC debugging to max, 60s are more like a blink of an eye.

vrurg: @niner It is a tradeoff. How many people do GC debugging? While there is an average limit for human's patience. 5-15 seconds range fits into the limit but 60 is definitely not. :) 

If we're gonna take things like this into account, we probably need to provide a default timeout value somehow which would depend on wether some heavy task is on or off. Then, say, `Promise.in($*TEST-TIMEOUT)` would handle such situations well enough.

niner:
> @niner It is a tradeoff. How many people do GC debugging? While there is an
> average limit for human's patience. 5-15 seconds range fits into the limit
> but 60 is definitely not. :) 

Who would need to wait? The bug is fixed. So the only case where one would 
even notice the delay is when one causes a regression. The chance for that is 
very small and even then the wait time would strike only once as the affected 
developer would probably start debugging right away.

At the same time, slowing rakudo down (even with just MVM_SPESH_BLOCKING and 
MVM_SPESH_DELAY) is something MoarVM developers frequently do and especially 
because of that, I tend to heavily load my machine with parallel tests, so it 
doesn't take any longer than really necessary.

vrurg:
@niner Get me right, I'm not against raising the timeout anyway. The thing I probably should've mentioned and which is the reason to propose something like `TEST-TIMEOUT` is that it is not the first test where this approach is used. So, in either case, I'd rather support providing some kind of standard value of minimal timeout to be used. Would ever somebody needs bigger one then it could be `TEST-TIMEOUT × $SOME-MULTIPLIER`.

My quick grepping over roast revealed that there is a number of tests utilizing the timeout checks, so probably my proposal would make a good sense to adjust them all. And if nobody minds, the value could be provided as a constant in `Test` module.

dumarchie:
The new test passes when the [fix](/rakudo/rakudo/pull/3447) is merged.

PR #602, opened 2019-12-20
Extend test for trailing documentation on sub params

kalkin:
In the current rakudo version 2019.11 this is broken. On the current
rakudo/rakudo master branch this passes. I extended the test, so we
shall never more break this functionality.


tbrowder:
Well that's good news!

PR #603, opened 2019-12-21
Test for basic behaviors of submethods in roles for 6.e

vrurg:
This aspect wasn't covered previously. Yes, my bad, I overlooked it...


vrurg:
Support for #3371

PR #625, opened 2020-02-26
Test new env var RAKULIB

tbrowder:
Test is for the new RAKULIB environment variable implemented in Rakudo PR #3516.


S11-modules/rakudolib.t
```diff
@@ -0,0 +1,15 @@
+use v6;
+use Test;
+use lib $?FILE.IO.parent(2).add("packages/Test-Helpers");
+use Test::Util;
+
+plan 1;
+
+# Test deprecated environment variable RAKUDOLIB
+# TODO remove this file when RAKUDOLIB is removed
+is_run 'BEGIN { BEGIN { q{S11modulesRakuLibTest.rakumod}.IO.spurt(q{package { say q{all your base} }}); %*ENV<RAKULIB>=qq{}; }; use S11modulesRakuLibTest }',
```
patrickbkr: Once it's in roast it's specced. We should not spec RAKUDOLIB. Especially not in roast (which is independent of a concrete implementation, such as rakudo).
So I'd like to have this test removed.

tbrowder: I'll remove the test. And I'll try to move it to Rakudo later if that's okay.

vrurg: Wait with moving until something concrete is decided in raku/problem-solving#167.

tbrowder: Okay, it's removed from roast, but I have done nothing else.

ugexe:
There really isn't a need to create/load a module -- we only need to check that the path is available for loading modules. For example:
```
$ perl6 -e 'BEGIN my $spec = "file#{q|./foobarbaz|.IO.absolute}"; say $*REPO.repo-chain.grep({ .path-spec eq $spec }).so;'
False

$ perl6 -e 'BEGIN my $spec = "file#{q|./foobarbaz|.IO.absolute}"; say $*REPO.repo-chain.grep({ .path-spec eq $spec }).so; BEGIN %*ENV<PERL6LIB> = $spec'
True
```

vrurg:
Is this one ready for merge?

@ugexe I think just testing for the path existing in the specs is still not enough. By actually loading something we may cover regressions in module loading code in general. In other words, I would treat this test as the ability to load a module from a specified path.

ugexe:
We have tons of tests, many implicit, that ensure a given path can load modules.

S11-modules/perl6lib.t
```diff
@@ -5,11 +5,13 @@ use Test::Util;
 
 plan 1;
 
-# TODO Replace PERL6LIB with RAKULIB when supported
-is_run 'BEGIN { BEGIN { q{S11modulesRakuLibTest.pm6}.IO.spurt(q{package { say q{all your base} }}); %*ENV<PERL6LIB>=qq{}; }; use S11modulesRakuLibTest }',
+# Test deprecated environment variable PERL6LIB
+# TODO remove this file when PERL6LIB is removed
+is_run 'BEGIN { BEGIN { q{S11modules-PERL6LIB-Test.rakumod}.IO.spurt(q{package { say q{all your base} }}); %*ENV<PERL6LIB>=qq{}; };\
+        use S11modules-PERL6LIB-Test }',
```
ugexe: Why is this broken onto a new line?

S11-modules/perl6lib.t
```diff
@@ -5,11 +5,13 @@ use Test::Util;
...
 {
     out    => "",
     status => * != 0,
-}, 'RT 130883 is fixed';
+}, 'RT 130883 is fixed, PERL6LIB works';
```
ugexe: I don't think co-opting a security test for PERL6LIB is the best thing to do

ugexe:
At the very least there is no reason to dynamically create a module to do this test. Here is a folder full of modules that can be loaded already in the roast to point to.

S11-modules/perl6lib.t
tbrowder: So one can see all the text in a terminal with non-folding lines.

S11-modules/perl6lib.t
tbrowder: That was already there before this PR started.

tbrowder:
@ugexe, your ideas are good. Why don't I close this PR and you fix the tests. All I've done is piggy back on the existing test for PERL6LIB and I didn't see the need for a wholesale rewrite of all the existing env var module path tests.

ugexe:
Here is an untested example. I'm sure you can get it to work if it doesn't already.
```
use Test;

# you might need to adjust .parent(...) based on where you save this test
my $path-to-dist = $?FILE.IO.parent(2).add("packages/Example/lib").absolute;

subtest 'RAKULIB' => {
    my $env = %*ENV;
    $env<RAKULIB> = $path-to-dist;

    ok so run :$env, :!out, :!err, $*EXECUTABLE.absolute, "-e", "use Example::A";
}

subtest 'PERL6LIB (deprecated?)' => {
    my $env = %*ENV;
    $env<PERL6LIB> = $path-to-dist;

    ok so run :$env, :!out, :!err, $*EXECUTABLE.absolute, "-e", "use Example::A";
}
```

tbrowder:
Okay, I'll try that tomorrow. Thanks.

tbrowder:
Closing this PR--will resubmit after I use @ugexe's suggestions.

PR #628, opened 2020-03-02
RAKULIB and PERL6LIB problem

tbrowder:
The recent addition of RAKULIB to the master branch has somehow caused some problems with roast test files added or modified for RAKULIB and PERL6LIB.

See details of the problem with gist https://gist.github.com/tbrowder/21c84659996a32b836f9b711463e2a7c

I have not tried to return to the original test formats before the changes, but I will try them in another PR.

### DIFF
--- a/S11-modules/rakulib.t
+++ b/S11-modules/rakulib.t
@@ -11,9 +11,9 @@ plan 1;
 
 my $path-to-dist = $?FILE.IO.parent($level).add("packages/Example/lib").absolute;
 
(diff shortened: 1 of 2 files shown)


tbrowder:
Okay, I'll try removing them.

tbrowder:
I removed the lines you mentioned but the failure still occurs. It sure looks to me like a Rakudo bug. Rakudo is looking for the experimental module and dies when it can't find it. Weird!

PR #633, opened 2020-04-02
Cover special case of wrapping first multi-candidate

vrurg:
Solves a problem described by @jnthn in Raku/problem-solving#170. When
the first multi-candidate of a method is wrapped on a parent class the
control is then returned to the `MethodDispatcher` causing the rest
multi-candidates to be skipped.


vrurg:
Covers rakudo/rakudo#3590

PR #635, opened 2020-04-13
Test usage of labels within nested loops

usev6:
... as reported in https://github.com/rakudo/rakudo/issues/3622.


usev6:
These tests don't pass yet. Should be merged once https://github.com/rakudo/rakudo/issues/3622 is fixed.

usev6:
For the record: one could probably get inspirations for more tests from https://github.com/Raku/roast/commit/9f418777bc.

lizmat:
Thanks again!

PR #636, opened 2020-04-22
Add tests for Windows specific Proc argument quoting

patrickbkr:
This PR adds tests for [Raku/problem-solving#20](https://github.com/Raku/problem-solving/issues/20).

### DIFF
--- a/S17-procasync/windows-arg-quoting.t
+++ b/S17-procasync/windows-arg-quoting.t
@@ -1,0 +1,57 @@
+use v6;
+use Test;
+
+=begin pod
+
+Sub process arguments need to be quoted according to a specific schema
+to be understood by most processes on Windows.
+This quoting is tested here.
+
+See L<https://github.com/Raku/problem-solving/issues/20>
+
+=end pod
+
+plan 21;
+
+if !$*DISTRO.is-win {
+    skip-rest "windows argument quoting can't be tested on non-Windows";
+    exit;
+};
+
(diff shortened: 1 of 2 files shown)


S17-procasync/windows-arg-quoting.t
```diff
@@ -0,0 +1,54 @@
...
+to be understood by most processes on Windows.
+This quoting is tested here.
+
+See L<https://github.com/Raku/problem-solving/issues/20>
+
+=end pod
+
+plan 21;
+
+if !$*DISTRO.is-win {
```
ugexe, line 16: What about Java on windows? (it *did* work at one time)

patrickbkr: Good catch! Would you disable these tests altogether on JVM or only the failing ones?

ugexe: If NativeCall doesn't work on JVM it is probably a moot point. Do you happen to have a hunch on which ones would might fail on JVM?

patrickbkr: My guess is, that all the `test-run-verbatim` tests will fail because the JVM will just quote the things anyways (because the thing we call is not a .bat file).

patrickbkr: Now all of these tests are skipped on the JVM. I guess there isn't much point in testing JVMs internal quoting behavior.

PR #641, opened 2020-05-10
Fix external dependency

### LINKED ISSUE #640: Roast depends on Rakudo when it shouldn't
Checking out roast and running `perl fudgeandrun $(sed 's/#.*//' spectest.data | grep '\.t')` using the system wide installed rakudo master yields the following test failures:
```
[ 1809s] Test Summary Report
[ 1809s] -------------------
[ 1809s] S03-metaops/misc.t                                       (Wstat: 256 Tests: 0 Failed: 0)
[ 1809s]   Non-zero exit status: 1
[ 1809s]   Parse errors: No plan found in TAP output
[ 1809s] S26-documentation/block-leading-user-format.t            (Wstat: 256 Tests: 0 Failed: 0)
[ 1809s]   Non-zero exit status: 1
[ 1809s]   Parse errors: No plan found in TAP output
[ 1809s] integration/advent2010-day03.t                           (Wstat: 512 Tests: 2 Failed: 2)
[ 1809s]   Failed tests:  1-2
[ 1809s]   Non-zero exit status: 2
[ 1809s]   Parse errors: Bad plan.  You planned 12 tests but ran 2.
[ 1809s] Files=1361, Tests=188669, 1777 wallclock secs (24.89 usr  2.58 sys + 2077.74 cusr 109.43 csys = 2214.64 CPU)
[ 1809s] Result: FAIL
```

`S03-metaops/misc.t` and `S26-documentation/block-leading-user-format.t` fail because they try to load a module called `Test::Helpers` which seems to be part of rakudo's tests. `integration/advent2010-day03.t` fails because it expects the files `Makefile CREDITS LICENSE README.md` which are clearly part of rakudo and not roast.

This is blocking automated spectests for every rakudo commit on the Open Build Service.

JJ:
By putting it in here. Closes #640.

### DIFF
--- a/S26-documentation/block-leading-user-format.t
+++ b/S26-documentation/block-leading-user-format.t
@@ -1,7 +1,5 @@
 use v6;
 use Test;
-use lib <t/packages>;
-use Test::Helpers;
 
 plan 2;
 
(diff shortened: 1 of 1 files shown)


tbrowder:
Looks good to me. Thanks, @JJ!

vrurg:
Actually, considering https://github.com/Raku/roast/blob/master/packages/Test-Helpers/lib/Test/Util.pm6#L107 – what's the point? I highly recommend reverting this PR and re-implementing it with the standard routine.

JJ:
OK, I'll do it adding to the other one, instead of reverting this commit.

JJ:
The problem is that the signature is not exactly the same, not to mention the name:

```
Cannot resolve caller is_run(Str:D, Str:D, :compiler-args(Array), :exitcode(Int), :out(Str)); none of these signatures match:
    (Str $code, %expected, *%o)
    (Str $code, Str $input, %expected, *%o)
    (Str $code, %expected, Str $name, *%o)
    (Str $code, Str $input, %expected, Str $name, *%o)
  in block  at S26-documentation/block-leading-user-format.t line 56
  in block <unit> at S26-documentation/block-leading-user-format.t line 50
```

I can try and adapt the call to one of these signatures, or move the routine to that file. I'll see what can be done.

JJ:
Also, is-run uses standard `run` while lexically Test::Utils' got its own `run`. So moving it there does not really work either. I don't think it's worth the while to reverse it, or even to move *this* is-run to Test::Helpers, but I'd be happy to hear other opinions.

vrurg:
I have committed a fix which makes use of roast `is_run`. The differences are rather subtle as except for different signature I never had a problem with `is_run`.

PR #652, opened 2020-07-15
Add more tests for gather/take

usev6:
All of these tests are dying with an UnwindException on the
JVM backend currently, but they are passing on MoarVM.

### DIFF
--- a/integration/gather-with-loops.t
+++ b/integration/gather-with-loops.t
@@ -1,0 +1,144 @@
+use v6;
+
+use Test;
+
+plan 21;
+
+# https://github.com/rakudo/rakudo/issues/3634
+
+## last in loop
+{
+    my $res = gather loop { take 42; last };
+    is-deeply $res, (42).Seq, 'gather/take with last in loop';
+
+    $res = gather while 1 { take 42; last };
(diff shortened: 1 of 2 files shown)


vrurg:
I would say that since they are known to fail it makes sense to fudge them.

usev6:
Yes, I could do that. (Just in case you missed it: I took the easy route and specified in spectest.data to only run the test file on Moar.)

vrurg:
I did miss it. But it looks somewhat weird because they're not moar specific in neither way. My approach would be to fudge them with `skip` for jvm. And add a canary test using `is_run` which would be fudged with `todo`. In this case if the bug is fixed we'll get a notification.

vrurg:
Thanks @usev6! I'm awaiting for my jvm backend to compile in order to test this PR. Will merge later if it passes.

usev6:
@vrurg Thanks! (I didn't want to generate that much extra work for you. My main reason for creating a PR was to get a second pair of eyes for the sanity of the tests.)

vrurg:
The "much work" as you name it is only about running a compiler and then a single test. :)

Ok, done. Once again - thanks!

PR #655, opened 2020-07-22
Test for empty string match with ignore-case and ignore-mark

patrickbkr:
Related to rakudo/rakudo#3815.


patrickbkr:
This was fixed in [MoarVM/MoarVM 393ae5799](https://github.com/MoarVM/MoarVM/commit/393ae5799365a3ebfbb8ad3db3f73573b02c1044). So the matching tests can be merged now.

PR #661, opened 2020-08-03
S26t12merge

tbrowder:


tbrowder:
I can't figure this out. @vrug will have to deal with it somehow. But I'm ready to help if I can.

JJ:
The problem is here: https://github.com/Raku/roast/pull/661/commits/5349faafafe37c3b456e0050c2b7fa5a33c5b467, when you merged with git pull instead of git pull --rebase to put your commits on top of those. You need to clone a fresh repo, and then copy your files on top to avoid that.

PR #665, opened 2020-08-09
Lathropd patch 2

lathropd:
Update some public-facing perl6 references

### DIFF
--- a/TODO
+++ b/TODO
@@ -5,7 +5,7 @@ by synopsis (when applicable) or put them in the OVERALL section at the top
 if they apply to the test suite in general.
 
 NOTE that references to a t/ directory are for the t/ directory in the mu
-repository found here:  https://github.com/perl6/mu
+repository found here:  https://github.com/raku/mu
 
 Please also read t/deprecated-syntax.pod for common mistakes in the test
 suite, and clear them when you update tests.
--- a/TODO
+++ b/TODO
@@ -5,7 +5,7 @@ by synopsis (when applicable) or put them in the OVERALL section at the top
 if they apply to the test suite in general.
 
 NOTE that references to a t/ directory are for the t/ directory in the mu
-repository found here:  https://github.com/raku/mu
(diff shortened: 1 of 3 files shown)


TODO
```diff
@@ -5,7 +5,7 @@ by synopsis (when applicable) or put them in the OVERALL section at the top
 if they apply to the test suite in general.
 
 NOTE that references to a t/ directory are for the t/ directory in the mu
-repository found here:  https://github.com/perl6/mu
+repository found here:  https://github.com/raku/mu
```
JJ, line 8: Maybe capitalize that?

TODO
```diff
@@ -16,7 +16,7 @@ commit message:
 [TODO]: Added more tasks for S02.
 
 This way, people working on the test suite can easily spot these commits
-in #perl6.
+in #raku.
```
JJ, line 19: Not totally clear what the hashtag is about. If it's a hashtag, in Twitter #rakulang is used. And apparently in Rosettacode it should be capitalized.

docs/announce/6.e.md
```diff
@@ -32,10 +32,10 @@ This ChangeLog is targeted towards language users, to help with preparation to u
 compilers supporting latest language version. Thus, it does not contain every minute
 change to the specification that occurred. Implementations wishing to ensure full
 compliance with the new version of the language specification should execute the
-test suite available at https://github.com/perl6/roast/ and examine any failing tests.
+test suite available at https://github.com/raku/roast/ and examine any failing tests.
```
JJ, line 36: Again, capitalizing is probably better. It's redirected, but still it's better to use like it should.

TODO
Altai-man: I think this is about irc channel, though.

docs/announce/6.e.md
lathropd: Fixed the capitalization. Thanks!

PR #666, opened 2020-08-09
Lathropd patch 3

lathropd:
Update a public facing perl6.org url in the 6.d announcement


lathropd:
accidentally included part of lathropd-patch-2, so that’s created and reverted

PR #674, opened 2020-09-10
Tests for the new coercion semantics

vrurg:
In support of rakudo/rakudo#3891


vrurg:
Based upon Raku/problem-solving#227

PR #677, opened 2020-10-02
Test `&`-sigiled variables in `where` clauses

### LINKED ISSUE #650: &sigiled Callable in where clause not covered
```
my &block = { False }; sub s($a where &block) { say $a }; s 'OK';
sub check { False }; sub s($a where &check) { say $a }; s 'OK'
```
Both cases are supported by Rakudo but are not checked for in the spec.
discussion: https://colabti.org/irclogger/irclogger_log/raku-dev?date=2020-06-26#l31

zoffixznet:
Closes #650

### DIFF
--- a/S12-subset/subtypes.t
+++ b/S12-subset/subtypes.t
@@ -3,7 +3,7 @@ use Test;
 use lib $?FILE.IO.parent(2).add("packages/Test-Helpers");
 use Test::Util;
 
-plan 90;
+plan 91;
 
 =begin description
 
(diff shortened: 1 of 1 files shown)


AlexDaniel (approved):
Looks good to me. I guess you could also test multi-dispatch based on &-sigilled where clauses, but IMO this PR is more than good enough.

AlexDaniel:
By the way, none of the tests seem to use a block (instead of a sub), which the original ticket explicitly mentioned. Feel free to add tests for that, if you feel like it.

S12-subset/subtypes.t
```diff
@@ -482,4 +482,90 @@ subtest 'postconstraints on variables in my (...)' => {
...
+group-of 5 => '`&`- sigiled variable be used in where' => {
+    my $wanted;
+    sub pos-match { $wanted = $^got; True  }
+    sub neg-match { $wanted = $^got; False }
+
+    group-of 4 => 'subset' => {
+        $wanted = Nil;
+
+        my subset PosSubset where &pos-match;
+        my subset NegSubset where &pos-match;
```
niner, line 495: Should this read `where &neg-match`?

zoffixznet:
All comments and corrections addressed in https://github.com/Raku/roast/pull/678

PR #691, opened 2020-10-23
Add and fix tests for multi-dispatching methods

vrurg:
In support for rakudo/rakudo#3977.

### DIFF
--- a/S12-methods/multi.t
+++ b/S12-methods/multi.t
@@ -39,7 +39,7 @@ is($foo.bar(5), 'Foo.bar() called with Int : 5', '... multi-method dispatched on
 is($foo.bar(4.2), 'Foo.bar() called with Numeric : 4.2', '... multi-method dispatched on Numeric');
 
 throws-like { $foo.baz() }, X::Multi::NoMatch,
-    "Call with wrong nnumber of args results in no matching candidate exception";
+    "Call with wrong number of args results in no matching candidate exception";
 
 class Foo2 {
     multi method a($d) {   #OK not used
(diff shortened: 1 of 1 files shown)


S12-methods/multi.t
```diff
@@ -38,10 +38,8 @@ is($foo.bar("Hello"), 'Foo.bar() called with Str : Hello', '... multi-method dis
 is($foo.bar(5), 'Foo.bar() called with Int : 5', '... multi-method dispatched on Int');
 is($foo.bar(4.2), 'Foo.bar() called with Numeric : 4.2', '... multi-method dispatched on Numeric');
 
-# https://github.com/Raku/old-issue-tracker/issues/1019
-#?rakudo todo 'RT #66006'
-try { EVAL '$foo.baz()' };
-ok ~$! ~~ /:i argument[s?]/, 'Call with wrong number of args should complain about args';
+throws-like { $foo.baz() }, X::Multi::NoMatch,
+    "Call with wrong nnumber of args results in no matching candidate exception";
```
lizmat, line 42: s/nnumber/number

S12-methods/multi.t
```diff
@@ -115,7 +113,7 @@ is Bar.new.a("not an Int"), 'Any-method in Foo';
     is $m.d( '7' ), 'string',  'dispatch to other role';
     is $m.d( 1.2 ), 'any',     'dispatch to the class with the roles';
 
-    my @multi_method = $m.^methods.grep({ ~$_ eq 'd' });
+    my @multi_method = $m.^methods.grep({ .gist eq 'd' });
```
lizmat, line 116: Why not `.name` instead of `.gist` ?

lizmat (changes requested):
Two small nits, otherwise ok for merging :-)

S12-methods/multi.t
vrurg: I was to tired to check it deeper, but basically you back up my yesterday thoughts. Will commit the change later today.

PR #693, opened 2020-10-27
:enc(utf16) assumes host byte order if there is no BOM, so test accordingly.

nwc10:
The test had always used 'sample-UTF-16LE.txt' when testing :enc(utf16), but
that is only appropriate on little endian systems. For big endian systems,
we need to test with 'sample-UTF-16BE.txt' instead.

### DIFF
--- a/S32-io/utf16.t
+++ b/S32-io/utf16.t
@@ -7,13 +7,22 @@ plan 39;
 # to test .encode, .decode, .slurp, .spurt. We also make sure .read reads the
 # file properly and that utf16 will get rid of a BOM at the start of the file.
 
+# "Raku's utf16 format specifier will use the endianness of host system when
+# encoding. When decoding it will look for a byte order mark and if it is there
+# use that to set the endianness. If there is no byte order mark it will assume
+# the file uses the same endianness as the host system."
+# Hence for testing a file without a BOM, we must pick the correct ordering:
+
+my $matching_utf16 = $*KERNEL.endian == LittleEndian
+    ?? 'sample-UTF-16LE.txt' !! 'sample-UTF-16BE.txt';
+
 # TODO:
 # * Check that BOM other places other than the start of the file get passed
 #   through with 'utf16'.
 # * Test .lines with both single and multiple lines
 my %enc =
     utf8    => <sample-UTF-8.txt>,
-    utf16   => <sample-UTF-16LE-bom.txt sample-UTF-16BE-bom.txt sample-UTF-16LE.txt>,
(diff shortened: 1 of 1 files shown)


S32-io/utf16.t
```diff
@@ -7,13 +7,22 @@ plan 39;
 # to test .encode, .decode, .slurp, .spurt. We also make sure .read reads the
 # file properly and that utf16 will get rid of a BOM at the start of the file.
 
+# "Raku's utf16 format specifier will use the endianness of host system when
+# encoding. When decoding it will look for a byte order mark and if it is there
+# use that to set the endianness. If there is no byte order mark it will assume
+# the file uses the same endianness as the host system."
+# Hence for testing a file without a BOM, we must pick the correct ordering:
+
+my $matching_utf16 = $*KERNEL.endian eq 'LittleEndian'
```
MasterDuke17, line 16: Wouldn't `$*KERNEL.endian ~~ LittleEndian` be a little better?

niner:
> Wouldn't `$*KERNEL.endian ~~ LittleEndian` be a little better?

Since Enums are Ints by default, I'd say `$*KERNEL.endian == LittleEndian`

MasterDuke17:
> Since Enums are Ints by default, I'd say `$*KERNEL.endian == LittleEndian`

`==` is used in src/core.c/Buf.pm6, so I guess that's the way to go.

nwc10:
> `==` is used in src/core.c/Buf.pm6, so I guess that's the way to go.

as best I can tell, that ony works there because it creates the Enum in its source:

```
# externalize the endian indicators
enum Endian (
  NativeEndian => nqp::box_i(nqp::const::BINARY_ENDIAN_NATIVE,Int),
  LittleEndian => nqp::box_i(nqp::const::BINARY_ENDIAN_LITTLE,Int),
  BigEndian    => nqp::box_i(nqp::const::BINARY_ENDIAN_BIG,Int),
);
```

So for the single ternary for this spectest, one could write:

```
$*KERNEL.endian == nqp::const::BINARY_ENDIAN_LITTLE
```

Is that the best plan?

MasterDuke17:
> So for the single ternary for this spectest, one could write:
> 
> ```
> $*KERNEL.endian == nqp::const::BINARY_ENDIAN_LITTLE
> ```
> 
> Is that the best plan?

`$*KERNEL.endian == LittleEndian` returns `True` on my system (and `False` for `BigEndian`). Don't see why you'd need the NQP code.

nwc10:
Not sure what I messed up when I tried to test it. It *does* work for me too once I get it right.

Revised branch pushed.

PR #709, opened 2020-12-08
Make test conform to the new coercion semantics and role invocation

vrurg:
In support of rakudo/rakudo#4096. Previously the invocation handler was unconditionally using a coercion method named after the target type. Now it uses the new coercion protocol. With this change `Str($*USER)` ends up as `IntStr` because no actual coercion is needed. Correspondingly, testing for the outcome of it to be a `Str` class makes more sense than to test for it to be exactly a `Str` instance.

### DIFF
--- a/S02-literals/allomorphic.t
+++ b/S02-literals/allomorphic.t
@@ -458,6 +458,6 @@ group-of 4 => '.comb on allomorphs uses Str variant' => {
 }
 
 # https://github.com/rakudo/rakudo/issues/3308
-is Str($*USER).^name, "Str", 'No leaking of guts types when coercing allomorph to Str';
+isa-ok Str($*USER), Str, 'No leaking of guts types when coercing allomorph to Str';
 
 # vim: expandtab shiftwidth=4
--- a/S13-overloading/typecasting-long.t
+++ b/S13-overloading/typecasting-long.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 32;
+plan 39;
 
 # L<S13/"Type Casting"/"method CALL-ME(**@slice) {...}">
 # basic tests to see if the methods overload correctly.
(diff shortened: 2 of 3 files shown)


lizmat:
hmmm... so how would one coerce an `IntStr` to a `Str` then ?

vrurg:
@lizmat explicitly: `$*USER.Str === Str`

lizmat:
So something like `Str(IntStr) $foo` in a signature, will *not* give you a `Str`?

vrurg:
Apparently not. And it must not because `IntStr ~~ Str`.

Leont:
That can be kind of annoying, given that an `IntStr` is not Liskov substitutable for an `Int`

vrurg:
@Leont Unfortunately, there is no way to formally explain why `Bar is Foo` is different from `Allomorph is Str`.

lizmat:
Having said that, it is at least *my* expectation that the result of a coercion to `Foo` would yield a `True` on `coercion-result =:= Foo`.

vrurg:
@lizmat consider this:

```
class Foo {...}
class Bar is Foo {...}
sub f1(Foo $v) {...}
sub f2(Foo() $v) {...}
f1(Bar.new);
f2(Bar.new);
```

Why should f2 fail in your case?

lizmat:
@vrurg Since `class A { }; class B is A { }; sub a(A() \b) { dd b }; a B` produces `B` since before the coercion protocol changes, I guess I will need to reprogram my expectations  :)

vrurg:
@lizmat But the subject of coercing exactly into the target type makes me considering a kind of forced coercion. Technically it's not hard to be implemented. But practically I don't see a good syntax form for it. So far, the best what comes into mind is `Str(IntStr):f` and it looks highly imperfect.

lizmat:
Yeah, adding syntax like that would require a problem-solving issue, I guess.

PR #713, opened 2020-12-30
Two more spectests that need tweaking to work on Solaris

nwc10:


lizmat:
I assume these can be merged before the Solaris work is actually merged?

nwc10:
To my mind the strict answer to "can be merged?" is independent of whether MoarVM copes with Solaris :-)

But

MoarVM does now cope with Solaris (ie changes to get it building, at least with gcc) are in, so merging these two changes reduces the number of failing spectests. (if one builds with MoarVM master)

PR #715, opened 2021-02-14
Add test for R-metaop as value in colon pair

atroxaper:
In support for [#4204](https://github.com/rakudo/rakudo/pull/4204)

### DIFF
--- a/S03-metaops/reverse.t
+++ b/S03-metaops/reverse.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 69;
+plan 71;
 
 =begin pod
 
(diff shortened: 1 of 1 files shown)


MasterDuke17:
I'd prefer tests that didn't depend on the string representation of something (especially the `.gist`, which is only meant as a human-readable representation). Maybe check that the value of `foo` is `1` instead?

atroxaper:
@MasterDuke17 I agree. I’m going to create a Capture and check equality.

vrurg:
@atroxaper Why not just use a named? `sub foo(:$foo) { note $foo }; foo(:foo(3 R- 2))` currently fails, so the test could be done as (note: quick and dirty example):

```
my $got;
sub foo(:$foo) { $got = $foo };
lives-ok { foo(:foo(3 R- 2)) }, "call doesn't throw";
is $got, -1, "named is good";
```

atroxaper:
@vrurg Did so. Thank you.

PR #717, opened 2021-02-20
Tests for rakudo PR #4208

salortiz:
Seems that the List creation via 'is List' wasn't tested.
As this is basic syntax I think tests are deserved in roast.
Can be merged after merging  https://github.com/rakudo/rakudo/pull/4208

### DIFF
--- a/S32-list/create.t
+++ b/S32-list/create.t
@@ -17,21 +17,20 @@ is($list_sub, (1, 2, 3), 'The &list() function created a list.');
 is(+$list_sub, 3, 'Finding the length of the list works as expected.');
 
 {
-my $list_obj = List.new(4, 5, 6);
-isa-ok($list_obj, List, 'Creating a new list object with new works.');
-is($list_obj, list(4, 5, 6), 'The list object contains the right values.');
-is(+$list_obj, 3, 'Finding the length functions properly.');
+  my $list_obj = List.new(4, 5, 6);
+  isa-ok($list_obj, List, 'Creating a new list object with new works.');
+  is($list_obj, list(4, 5, 6), 'The list object contains the right values.');
+  is(+$list_obj, 3, 'Finding the length functions properly.');
 }
 {
-my @list-pos is List = 1, 2, 3;
-isa-ok(@list-pos, List, "List can be created by 'is List'");
-isa-ok(@list-pos, Positional, 'Is Positional');
-is(@list-pos, (1, 2, 3), 'The list objects contains the right values.');
+  my @list-pos is List = 1, 2, 3;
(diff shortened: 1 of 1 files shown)


S32-list/create.t
```diff
@@ -22,7 +22,17 @@ isa-ok($list_obj, List, 'Creating a new list object with new works.');
 is($list_obj, list(4, 5, 6), 'The list object contains the right values.');
 is(+$list_obj, 3, 'Finding the length functions properly.');
 }
-
+{
+my @list-pos is List = 1, 2, 3;
```
vrurg, line 26: I see the reason why you don't indent here, but to my view it's either of two would be better options:

- indent the old test code
- don't use blocks

Considering that you're referring to an existing PR, indenting the blocks is preferable if accompanied with a comment referring to the PR page. Something like:

```raku
# https://github.com/rakudo/rakudo/pull/4208
{
    ...; # test code
}
```

If the PR, in turn, fixes an existing ticket, then a link to the ticket would be a better choice.

S32-list/create.t
```diff
@@ -22,7 +22,17 @@ isa-ok($list_obj, List, 'Creating a new list object with new works.');
...
+isa-ok(@list-pos, List, "List can be created by 'is List'");
+isa-ok(@list-pos, Positional, 'Is Positional');
```
vrurg, line 28: I think this test is redundant. `List` is a `Positional` by definition. If there be a test for this assertion then most definitely belongs to another test suite.

S32-list/create.t
```diff
@@ -22,7 +22,17 @@ isa-ok($list_obj, List, 'Creating a new list object with new works.');
...
+is(@list-pos, (1, 2, 3), 'The list objects contains the right values.');
+}
+{
+my @list is List = 'foo';
+is(@list, ('foo',), 'The list constain a single element');
+is(+@list, 1, 'The lenght is one');
```
vrurg, line 34: Just a minor note on test logic and descriptions. The primary purpose of the test is to ensure the list is getting a single element. So, it makes sense to test for the number of elements first. And then test and explain if the element value what we expect it to be, not that it's just "a single element".

vrurg (changes requested):

vrurg (approved):

S32-list/create.t
salortiz: `List` is a `Positional` by definition, but that "definition" isn't yet tested in roast.
Any way, indeed I think that test belongs to another file. I'll search for a better place.

patrickbkr:
Any reason not to merge this now that the fix is in?

lizmat:
Nope, thanks for the nudge!

PR #720, opened 2021-03-01
Mark utf32 tests as todo for rakudo.moar

codesections:
Rakudo.moar was previously passing a test for utf32 support despite not actually supporting utf32.  This updates the test to more
reliably test for utf32 support and marks the test as todo for the MoarVM backend.

Without this change, rakudo/rakudo#4232 fails Roast spectests.

Related issues:  rakudo/rakudo#3293 and MoarVM/MoarVM#1348


lizmat:
Thanks again!

PR #725, opened 2021-03-30
Changes to test new predicate `wait`

JJ:
There's another test for non-predicate `wait`, so we're good here. Refs (and maybe solves) rakudo/rakudo#4275

### DIFF
--- a/S17-lowlevel/lock.t
+++ b/S17-lowlevel/lock.t
@@ -97,17 +97,17 @@ plan 23;
         my $c = $l.condition;
         my $now1;
         my $now2;
-        my $counter = 0;
+        my $flag = False;
         my $t1 = Thread.start({
             $l.protect({
-                $c.wait( { $counter != 0 } );
+                $c.wait( { $flag } );
                 $now1 = now;
             });
         });
 
         my $t2 = Thread.start({
             $l.protect({
-                $counter++;
+                $flag = True;
                 $c.signal();
             });
         });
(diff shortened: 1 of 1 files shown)


S17-lowlevel/lock.t
```diff
@@ -100,9 +100,7 @@ plan 23;
         my $counter = 0;
         my $t1 = Thread.start({
             $l.protect({
-                while $counter == 0 {
```
vrurg, line 103: But we haven't removed this functionality, only added new. So, perhaps we need a new test?

JJ: That's tested a bit down the line.

vrurg: Yes, but it's a bit different scenario down there. Because `wait` is a multi, there is very tiny, but still a non-zero, chance that something could break. For example, re-entering the method would break whereas the predicate version doesn't re-enter and would be immune to the problem.

JJ:
Ok, will do

S17-lowlevel/lock.t
JJ: The main problem I see here is that this code is actually written to *not* run. So we're not actually testing it. Let me see what can be done.

JJ: `.wait()` is also tested in line 68 in the same file. Is that what you mean? That's what I meant in the PR body. There were two instances of wait(); I've turned one of them to test the other multi.

vrurg: Actually, I meant line 150, but that's irrelevant. My point is that neither 68 nor 150 are testing for repetitive invocations of `wait` withing a lock. And your PR actually removes the only case where the case is tested.

JJ:
Ok, I get it now. Thanks!

vrurg (approved):

PR #726, opened 2021-03-31
Improve test for subtest output

vrurg:
Take into account additional header from a subtest output which reports
subtest message and looks like:

   # Subtest: message

Also improve testing for indented lines. If we ever decide to change the
number of spaces in subtest indentation then we won't need to change the
test. It now allows any number of spaces but only ensures the number
doesn't change across subtest output.

Support for rakudo/rakudo#4280 and rakudo/rakudo#4266

### DIFF
--- a/S24-testing/3-output.t
+++ b/S24-testing/3-output.t
@@ -51,8 +51,24 @@ is_run ｢
         ok 1, "foo\nbar";
         subtest "meow" => { ok 1, "mass\nbass\nmiss\nbiss" }
     ｣,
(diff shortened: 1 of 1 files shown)


vrurg:
@Altai-man I need your opinion about possible impact of this change on `-errata` branches.

Altai-man:
>I need your opinion about possible impact of this change on -errata branches.

This is a simple matter to me: if it passes or not. Adding another `#` comment like you did _should_ be safe IMO. If we test our TAP emitter anywhere, these tests should probably be updated.

vrurg:
Ok, this should be a cherry-pick candidate for -errata, perhaps.

@Altai-man thanks!

PR #729, opened 2021-04-06
Add notes for tests to fix for making method posix a Real number

tbrowder:
One proposed change to Rakudo (Rakudo issue #4286) is to make class DateTime method 'posix' return a Real number rather than an Int. During a first attempt at a simple change to do that, 17 of over 300 existing tests were found to fail. Those tests were annotated in the two failing test files for reference during future attempts.

This PR will add those TODO annotations to roast.


tbrowder:
Unable to resolve conflicts here. Will resubmit later.

PR #736, opened 2021-06-16
Test cases where mkdir should fail

McSinyx:
The added tests should pass for MoarVM/MoarVM#1507.  Edit: the final diff is confusing, please consider viewing each commit for the actual changes.

### DIFF
--- a/S32-io/mkdir_rmdir.t
+++ b/S32-io/mkdir_rmdir.t
@@ -3,7 +3,7 @@ use Test;
 use lib $?FILE.IO.parent(2).add("packages/Test-Helpers");
 use Test::Util;
 
-plan 5;
+plan 6;
 
 # Tests for IO::Path.mkdir and IO::Path.rmdir
 #
@@ -14,6 +14,13 @@ plan 5;
     my $dir = make-temp-dir;
     ok $dir.e, "$dir exists";
     ok $dir.d, "$dir is a directory";
+}
+
+# mkdir soft-fails when pathname exists and is not a directory.
+{
+    my $file = make-temp-dir.add: "file";
+    spurt $file, "hello world";
+    fails-like { $file.mkdir }, X::IO::Mkdir;
 }
 
 # rmdir soft-fails when dir doesn't exist.
--- a/S32-io/mkdir_rmdir.t
+++ b/S32-io/mkdir_rmdir.t
@@ -3,7 +3,7 @@ use Test;
 use lib $?FILE.IO.parent(2).add("packages/Test-Helpers");
 use Test::Util;
 
-plan 7;
+plan 5;
 
 # Tests for IO::Path.mkdir and IO::Path.rmdir
 #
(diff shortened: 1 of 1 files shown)


S32-io/mkdir_rmdir.t
```diff
@@ -23,6 +23,12 @@ plan 6;
     fails-like { $file.mkdir }, X::IO::Mkdir;
 }
 
+# mkdir soft-fails when directory name is too long (over 255 bytes).
+{
+    my $path = make-temp-dir.add("foobar").add("universe" x 42);
+    fails-like { $path.mkdir }, X::IO::Mkdir;
+}
```
zhuomingliang, line 30: I think this test only pass on posix, on windows it support long path.

niner:
Windows imposes a limit of 260 bytes on the total path length.

The maximum length of a path component (i.e. file or directory name) is file 
system dependent. Remember that the original FAT only allowed for 8+3 
characters, regardless of operating system. 255 bytes is a common limit, but 
by far not universal.

POSIX defines a PATH_MAX constant and on Linux it seems to be 4096 bytes:
linux/limits.h:13:#define PATH_MAX        4096  /* # chars in a path name 
including nul */

However, it looks like file systems will nevertheless allow for creating paths 
longer than that.

McSinyx:
> Windows imposes a limit of 260 bytes on the total path length.

There seems to be *Long Paths* on Windows, however I have yet to find how
it can affect components (which is also limited to 255 on NTFS by default):
https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation

McSinyx:
I'm not sure if the comment was clear enough, but this test case try
to trigger the case where the directory name (not the entire path name)
is longer than what is permitted, so that we can ensure that after
a successful creation, errors won't be silenced.  It could be other
reason but the name length seems to be the most portable and common case.

zhuomingliang:
+1 from me

vrurg:
@niner It looks like we have no legal way to find out what the value of `PATH_MAX` is and I'd say it's important enough to be available in a non-platform, non-backend dependable way.

McSinyx:
@vrurg, I think you missed my comment of it is about the directory/filename length limit, not total path name.  Though, I want to bring up that `mkdir` is inherently platform dependent, e.g. the file permission concept on Windows is quite different from the three-byte POSIX one `mkdir` on Raku uses.  In other words, if Reiser (some of the few [file systems known by Wikipedia to have filename length limit greater than 255](https://en.wikipedia.org/wiki/Comparison_of_file_systems#Limits)) is supported we might want a special indicator in the test case.

McSinyx:
> the test is dependent on information not directly known to it. It is assumed [...] To make it really reliable it must look like: [...] `$*VM.constants<PATH_MAX>`

That's true.  However, having such constant correctly defined, i.e. specifically for each file system (instead of just hardcoding to a number with one or two known exceptions) is nontrivial.  In addition, directory creations used by this test suite is making other assumptions as well, e.g. `make-temp-file` relies on PATH_MAX to be at least some certain number due to the inclusion of some prefixes and timestamps.  Therefore, I wonder if it is OK to not let perfect be the enemy of good here and allow an assertion that probabilistically always do the right job and a patch to MoarVM that (IMHO) correct the semantic that has been broken for at least half a decade.  Of course the opposite could have been said, that if it stayed broken for so long then it's probably OK to leave it that way, but I'd much prefer the other outcome.

vrurg:
I did forget about different file systems on the same OS. I also forget about all the peculiarities around `PATH_MAX`, but [this blog](https://insanecoding.blogspot.com/2007/11/pathmax-simply-isnt.html) reminded me of them.

Yet, it changes little. The chances are non-zero that the test will fail. For this reason we can't have it in the roast.

But what we can do is to move it into Rakudo test suite because it covers an implementation bug. This would make it easier to fix the test when necessary.

McSinyx:
I've moved the test to Rakudo as suggested.

vrurg (approved):

zhuomingliang (approved):

PR #740, opened 2021-06-23
Align test #28 of S26-documentation/02-paragraph.t with intended design of E<>

raydiak:
Lowercase strings in E<> should be valid HTML5 entities. The corrected test will fail without the corresponding Rakudo PR, so please merge rakudo/rakudo#4427 at the same time as this one. The fix could not be easily implemented for JVM due to Raku/nqp#727, so this also fudges the test under JVM


lizmat:
Thanks!

PR #742, opened 2021-06-28
Add test file for the 'modf' routine in rakudo

tbrowder:
See rakudo PR #4434.

### DIFF
--- a/S32-num/modf.t
+++ b/S32-num/modf.t
@@ -1,0 +1,91 @@
+use Test;
+
+# L<S32::Numeric/Numeric/"=item modf">
+
+plan 240;
+
+my %n =
+    '-123457e-3' => [-123, -0.457,            3],
+    '-10'        => [-10,   0,                0],
+    '-0x10'      => [-16,   0,                0],
+    '-0o10'      => [-8,    0,                0],
+    '-0xa'       => [-10,   0,                0],
+    '-0o127'     => [-87,   0,                0],
+    '-100'       => [-100,  0,                0],
+    '-5.9'       => [-5,   -0.9,              1],
+    '-5.499'     => [-5,   -0.499,            3],
+    '-2'         => [-2,    0,                0],
+    '-3/2'       => [-1,   -0.5,              1],
+    '-1.5e0'     => [-1,   -0.5,              1],
+    '-1.4999'    => [-1,   -0.4999,           4],
+    '-1.23456'   => [-1,   -0.23456,          5],
+    '-1'         => [-1,    0,                0],
+    '-0.5'       => [ 0,   -0.5,              1],
+    '-0.499'     => [ 0,   -0.499,            3],
+    '-0.1'       => [ 0,   -0.1,              1],
+    '-0'         => [ 0,    0,                0],
+    '-1.5e-10'   => [ 0,   -0.000_000_000_15, 11], # -0.000_000_000_15
+;
+
+
+# test the routine
+for %n.kv -> $val is copy, $arr {
+    my $int-part  = @($arr)[0];
+    my $frac-part = @($arr)[1];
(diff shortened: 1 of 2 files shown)


tbrowder:
Note there will be more tests added.

S32-num/modf.t
```diff
@@ -0,0 +1,91 @@
...
+    '-0.5'       => [ 0,   -0.5,              1],
+    '-0.499'     => [ 0,   -0.499,            3],
+    '-0.1'       => [ 0,   -0.1,              1],
+    '-0'         => [ 0,    0,                0],
+    '-1.5e-10'   => [ 0,   -0.000_000_000_15, 11], # -0.000_000_000_15
+;
+
+
+# test the routine
+for %n.kv -> $val is copy, $arr {
```
vrurg, line 32: Just an advise. You can shorten the code here by de-structuring the `$arr` in place:

```raku
for %n.kv -> $val is copy, ($int-part, $frac-part, $places) {
```

Aside of this, unless you want to intentionally make the test random-ordered, I consider use of hash redundant here. To my view, 

```raku
my @n = [ '-5.9', -5, -0.9, 1],
        ['-123457e-3', -123, -0.457, 3],
...

for @n -> ($val is copy, $int-part, $frac-part, $places) {
```

makes the code even more readable.

vrurg (changes requested):
I wonder if there're any chance for errors on user side. I'd be glad to see testing for incorrect user input. How `modf` must react on `*`? `Inf`, `NaN`?

Also, I didn't follow the original discussion about `modf`, but why `$val` is in stringified form only? This way you test the `Cool` candidate only. This is right, in the light of the proposed PR you kinda test every multi-candidate then. But testing must not rely on particular implementation.

And then, tests for positives must be explicitl, i.e. included into the table. When you use `abs` you turn `$val` into a `Numeric`, thus testing different multi-candidates compared to the tests for negative. It effectively means that you kind of testing two things at once: multi-dispatch and the type of argument.

S32-num/modf.t
tbrowder: @vrurg, I changed this to draft earlier because I'm still working on the Rakudo PR as well as the roast test code. I appreciate the comments, though, and you will probably see them used in the final results.

vrurg: Ok, will try to remember keeping my eye on this. Anyway, thank you for PRs!

tbrowder: You're very welcome, and I 'm learning a lot from this "exercise!"

tbrowder:
I am closing this PR because I will be rewriting it along with its associated Rakudo PR #4434 (which I have also closed).

tbrowder:
> I wonder if there're any chance for errors on user side. I'd be glad to see testing for incorrect user input. How `modf` must react on `*`? `Inf`, `NaN`?
> 
> Also, I didn't follow the original discussion about `modf`, but why `$val` is in stringified form only? This way you test the `Cool` candidate only. This is right, in the light of the proposed PR you kinda test every multi-candidate then. But testing must not rely on particular implementation.
> 
> And then, tests for positives must be explicitl, i.e. included into the table. When you use `abs` you turn `$val` into a `Numeric`, thus testing different multi-candidates compared to the tests for negative. It effectively means that you kind of testing two things at once: multi-dispatch and the type of argument.

I've been doing some experiments and would appreciate your thoughts on this:

+ use only one set of desired test values, all as strings with a leading '-' sign:

    my $xs = '-0.e5';

+ remove the leading minus signs from the strings to get positive versions:

    my $xpos = ($x ~~ s/^'-'//);

+ convert each to a RatStr or NumStr by using <<>> quotes, e.g., 'my $xr = <<$xs>>'

bGiven a string representation of all the negative test values

PR #746, opened 2021-07-06
Add test for #1486

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/1486

I really don't know where to put this, in `literals` there is nothing for hash literals and `hash` or `hashes` have nothing related also, instead testing just methods.

### DIFF
--- a/S09-hashes/objecthash.t
+++ b/S09-hashes/objecthash.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 55;
+plan 56;
 
 {
     class A { method Str() { 'foo' } };
@@ -138,6 +138,11 @@ plan 55;
 
 {
     eval-lives-ok 'my %*a{Int}', "Accept dynamic object hash"
+}
+
+# https://github.com/rakudo/rakudo/issues/1486
+{
+    ok {:foo, (if 0.5.rand < 1 { :bar })}, 'a conditional inside circumfix {} works';
 }
 
 #vim: ft=perl6


S09-hashes/objecthash.t
```diff
@@ -140,6 +140,11 @@ plan 55;
     eval-lives-ok 'my %*a{Int}', "Accept dynamic object hash"
 }
 
+# https://github.com/rakudo/rakudo/issues/1486
+{
+    ok {:foo, (if 1 { :bar })}, 'a conditional inside circumfix {} works';
```
lizmat, line 145: Are we sure the static optimizer doesn't optimize the `if 1` away?

Altai-man: I hope the new version is complex enough?

S09-hashes/objecthash.t
```diff
@@ -140,6 +140,11 @@ plan 55;
     eval-lives-ok 'my %*a{Int}', "Accept dynamic object hash"
 }
 
+# https://github.com/rakudo/rakudo/issues/1486
+{
+    ok {:foo, (if 0.5.rand < 1 { :bar })}, 'a conditional inside circumfix {} works';
```
vrurg, line 145: I'd like to ask about a comment explaining the approach about `if`. It's going to be really confusing what purpose of it without the background of this PR.

BTW, a sufficiently sophisticated static optimizer would optimize away this `if` too. But, unfortunately, we're not going get one of the kind any time soon.

vrurg (changes requested):

lizmat:
Closing as the problem is more complex

PR #747, opened 2021-07-07
Add test for #2920

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/2920

### DIFF
--- a/S03-sequence/misc.t
+++ b/S03-sequence/misc.t
@@ -1,6 +1,6 @@
 use v6;
 use Test;
-plan 31;
+plan 32;
 
 is ("fom" ... /foo/), "fom fon foo", "can use regex for endpoint without it being confused for closure";
 
@@ -130,5 +130,10 @@ is (1, { $^n*2 + 1 } ... 31, *+5 ... { $^n**2 > 2000 }, 'a', *~'z' ... { $_.char
 
 is-deeply (<a b c>, *.reverse ... *)[5], <c b a>.Seq,
     '… op does not prematurely cause consumtion of Seqs';
+
+# https://github.com/rakudo/rakudo/issues/2920
+{
+    lives-ok { (0, { $_ == 1000 ?? die ‘ouch!’ !! $_ + 1 } ... 99999).gist }, 'Large non-Infinite gists are properly detected as lazy';
+}
 
 # vim: expandtab shiftwidth=4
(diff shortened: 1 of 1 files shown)


S03-sequence/misc.t
```diff
@@ -131,4 +131,9 @@ is (1, { $^n*2 + 1 } ... 31, *+5 ... { $^n**2 > 2000 }, 'a', *~'z' ... { $_.char
 is-deeply (<a b c>, *.reverse ... *)[5], <c b a>.Seq,
     '… op does not prematurely cause consumtion of Seqs';
 
+# https://github.com/rakudo/rakudo/issues/2920
+{
+    lives-ok { (0, { $_ == 1000 ?? die ‘ouch!’ !! $_ + 1 } ... 99999).gist }, 'Large non-Infinite gists are properly detected as lazy';
```
vrurg, line 136: I'm trying to reproduce the bug and see if there is any better approach to test for it. With this test it looks to me as we're implying a limit on how many items `gist` must iterate over and it doesn't look good to me. This kind of thing should be implementation specific.

It feels to me that the correct solution must be in direct check if `iz-lazy` gives the right answer.

Altai-man: But what if `is-lazy` lies? Hmm, we _probably_ have test for this case. I still think that the test is fine as it is. If one for some reason bumps the limit of items displayed in a gist, this test will just fail immediately and we can tweak if if the time comes.

vrurg: I have finally managed to find the commit which fails and done a bit of testing. It looks like `gist` is just _a_ case where it dies. Using `is-lazy` on the same `Seq` results in the same outcome. The return value of it, apparently, is not relevant then. Yet, I think `is-lazy` is preferable over `gist` because the latter obscures the cause. And it does so in two ways. The first one is as in my case: it causes one to think that the throw is a result of `gist` evaluating over the list. The second is the fact that currently `is-lazy` returns `False` on this sequence, and yet the original test dies. This also makes the test message misleading: the sequence is not properly detected as lazy anyway, not on the latest master, and I'm not sure if it must be. Yet, it dies on that commit.

Altai-man: Sorry to waste your time like this, you are correct. I'll patch the test right away, thank you for taking your time and expertise to check the PR!

vrurg:
Great, thanks!

PR #750, opened 2021-07-17
Track changes caused by NQP PR #732

MasterDuke17:
The TODO is now passing, and the new exception thrown is a better one.

Should only be merged if/when https://github.com/Raku/nqp/pull/732 is merged and Rakudo is bumped.


JJ (approved):
Looks fine! Thanks.

PR #758, opened 2021-10-04
Simplify the code used to test destructors

MasterDuke17:
Sometimes under heavy load we'd hit the timeout instead of a GC
triggering the destructor. So instead, just loop until we hit it, but
add in a call to explicitly request a GC (which wasn't available when
this test was written).


vrurg (approved):
Deterministic test? Yes, for sure!

PR #786, opened 2022-01-22
Tweak some uint tests

usev6:
This shouldn't change any functionality (except for fudging some tests
for the JVM backend).

### DIFF
--- a/S02-types/int-uint.t
+++ b/S02-types/int-uint.t
@@ -24,7 +24,7 @@ plan 11 * @inttypes + 4;
 for @inttypes -> $type {
     my ($minval,$maxval) = ::($type).Range.int-bounds;
 
-    # TODO: merge this if/else into one test once the fundge isn't needed
+    # TODO: merge this if/else into one test once the fudge isn't needed
     if $type eq "uint64" {
         #?rakudo.jvm todo 'getting -1 instead of 18446744073709551615'
         is EVAL("my $type \$var = $maxval; \$var"), $maxval,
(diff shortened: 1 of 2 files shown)


S02-types/int-uint.t
```diff
@@ -59,25 +57,17 @@ for @inttypes -> $type {
...
-              Exception,
-              "setting $type to more than $maxval throws";
-        }
-        else {
-            #?rakudo todo 'setting more than max throws'
-            throws-like { EVAL "my $type \$var = {$maxval+1}" },
-              Exception,
-              "setting $type to more than $maxval throws";
-        }
+        #?rakudo todo 'setting more than max throws'
```
MasterDuke17, line 67: Don't these pass for the MoarVM backend now?

usev6: I don't think so, for example:
```
$ ./rakudo-m -e 'my int8 $var = 128; dd $var'
-128
```

Only for the ```int64``` and ```uint64``` cases there is an exception:
```
$ ./rakudo-m -e 'my $maxval = (::int64).Range.int-bounds[1]; say $maxval; my int64 $var = $maxval + 1'
9223372036854775807
Cannot unbox 64 bit wide bigint into native integer
  in block <unit> at -e line 1
$ ./rakudo-m -e 'my $maxval = (::uint64).Range.int-bounds[1]; say $maxval; my uint64 $var = $maxval + 1'
18446744073709551615
Cannot unbox 65 bit wide bigint into native integer
  in block <unit> at -e line 1
```

MasterDuke17 (approved):

PR #791, opened 2022-02-06
Try to avoid generating a core dump in segfault Proc test

niner:
The test intentionally triggers a segfault. Especially on developer's systems, this will create a core dump that will be listed by e.g. `coredumpctl list` which makes it harder to notice actual bug-related segfaults. We can avoid this by disabling core dumps for the process. Unfortunately there doesn't seem to be a portable way to do this. The only way to do this without changing the nature of the test (by e.g. running the process through a shell) is to use NativeCall again to call `setrlimit` On the bright side, if the setrlimit call fails (because e.g. it's not available on Windows), we do not need to care, as worst case we revert to the previous behaviour.


MasterDuke17 (approved):

MasterDuke17:
This does prevent the actual core file from being created, but there is still an entry in `coredumpctl list`. But what about faking the segfault entirely? E.g.,
```raku
my $p = Proc.new;
use nqp;
nqp::bindattr(nqp::decont($p), Proc, q|$!exitcode|, 0); nqp::bindattr(nqp::decont($p), Proc, q|$!signal|, 11);
use Test;
throws-like { sink $p }, X::Proc::Unsuccessful;
```
This passes for me.

niner:
I think I tried that and it still created a coredump

MasterDuke17:
> I think I tried that and it still created a coredump

Huh, doesn't for me (using arch linux).

PR #792, opened 2022-02-08
[JVM] Fudge some exploding tests

usev6:


usev6:
Not needed anymore -- see https://github.com/rakudo/rakudo/issues/4732.

PR #802, opened 2022-02-23
Change .parent behavior to "stupid" resolving

patrickbkr:
This seems to better capture the original intent when pondering the design
docs.


niner (approved):
Matches my reading of the design docs

ugexe (approved):

PR #810, opened 2022-06-16
Make spectest pass on Solaris

vlmarek:


vlmarek:
I wanted to mention the ksh reproducible case - $PWD is `/` even though we are in `/tmp`

```
$ bash -c "set -x; cd /tmp; PWD=/ /bin/ksh -c 'echo \$PWD'"
+ cd /tmp
+ PWD=/
+ /bin/ksh -c 'echo $PWD'
/
```

S29-os/system.t
```diff
@@ -103,10 +103,28 @@ throws-like { shell("program_that_does_not_exist_ignore_errors_please.exe") },
...
-                                    !! 'echo $PWD';
+
+    my @run-cmd;
+    my $shell-cmd;
+
+    given $*DISTRO {
+        when .is-win {
+            @run-cmd  = ('cmd.exe', '/C', 'echo %CD%');
+            $shell-cmd = 'echo %CD%';
+            say "win";
```
niner, line 114: looks like left over debug output?

S29-os/system.t
```diff
@@ -103,10 +103,28 @@ throws-like { shell("program_that_does_not_exist_ignore_errors_please.exe") },
...
+        }
+        when .Str.starts-with("solaris") {
+            # Solaris has ksh as /bin/sh which changes $PWD env var only when
+            # builtin 'cd' is called
+            @run-cmd  = ('/bin/sh', '-c', '/usr/bin/pwd');
+            $shell-cmd = '/usr/bin/pwd';
+            say "solaris";
```
niner, line 121: same here

niner (changes requested):

S29-os/system.t
vlmarek: Yup, sorry, thanks for spotting that

S16-io/eof.t
```diff
@@ -79,7 +79,7 @@ subtest '.eof on empty files' => {
         .read: 42;
         is-deeply .eof, True,  'eof is True after a read';
     }
-    with "/proc/$*PID/status".IO -> $p {
+    with "/proc/$*PID/environ".IO -> $p {
```
vrurg, line 82: I doubt the sanity of this test which is currently linux-like-centric. Neither `status` nor `environ` work for FreeBSD, for example. _/proc_ as such doesn't work neither on macos nor on Windows, etc. Guess, the file is to be chosen on platform-based grounds for better coverage.

- FreeBSD has `/proc/*/status`, as well as Linux.
- _environ_ would serve for Solaris then
- I'm not sure of the purpose of the test is to check for a special file accessibility, but macos has nothing but devfs mounted. So, _/etc/profile_ would be the best candidate after all
- Neither special mounts exists on Windows. So, _/Windows/win.ini_ or _system.init_ perhaps?

vlmarek:
So how about creating temporary file? Or using the one created few lines
above? Just fill it with some data. That should conclude whether .eof works
or not and should be platform independent, right? I'm happy to modify my
change if it is acceptable.

S16-io/eof.t
vlmarek: All right, so how about making the /proc/*/environ solaris specific only?

vrurg: I'm only for it. And skip on macos and win*.

vlmarek:
Ok, I'm confused now, sorry :) Does my current change look ok, or do we want to do it any other way? Thank you

vrurg (approved):

PR #816, opened 2022-08-15
Add more tests for sigilless declarations

vrurg:
Related to rakudo/rakudo#3919 and rakudo/rakudo#5027


vrurg:
Requires rakudo/rakudo#5030

PR #821, opened 2022-09-07
add lives-ok test of partially dimensioned view

grondilu:
add minimal test for partially dimensioned views of fixed-sized array, which are NYI at the moment I think.

In relation with https://github.com/Raku/old-issue-tracker/issues/6075

### DIFF
--- a/S09-multidim/indexing.t
+++ b/S09-multidim/indexing.t
@@ -1,7 +1,7 @@
 use v6;
 use Test;
 
-plan 31;
+plan 32;
 
 my @arr := Array.new(:shape(2;2));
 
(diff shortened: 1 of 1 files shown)


lizmat:
Shouldn't that be TODOd ?

grondilu:
> Shouldn't that be TODOd ?

I have no idea.

lizmat:
It currently fails, does it not?

grondilu:
It does fail since it's NYI.  I thought that shouldn't stop it from being specced, but maybe I was wrong.

lizmat:
Indeed, it shouldn't stop it from being specced.  But then it should be marked as "todo" by adding a line such as:

````
#?rakudo todo "partially dimensioned NYI"
````

**before** it  :-)

grondilu:
ok, I guess I need to read more about how this repo works.  Closing for now.

lizmat:
Why close it???

vrurg:
@grondilu Don't be so easily scared off! :) Check README.md for `fudge` preprocessor and its syntax. Find some examples in other tests by grepping for `#?rakudo`. That's it.

Good luck! More tests is less errors in the compiler, apparently. ;)

PR #822, opened 2022-09-16
Add tests for picking ver/api/auth default from META6

vrurg:
In support of rakudo/rakudo#5060

### DIFF
--- a/packages/S11-modules/multi-version/lib/MultiVer.rakumod
+++ b/packages/S11-modules/multi-version/lib/MultiVer.rakumod
@@ -1,0 +1,142 @@
+use v6;
+unit class MultiVer;
+use Test::Util;
+
+class Foo:ver<0.0.2> { }
+
+has $.prefix;
+has @.distributions;
+has $.standalone;
+has $.standalone-script;
+has $.inst-dir;
+has $.inst-repo;
+has $.use-lib;
+
+submethod TWEAK {
+    $!prefix //= make-temp-dir;
+}
+
+method mksubdir(IO:D() $path) {
+    my $fullpath = $path.is-relative ?? $!prefix.add($path) !! $path;
+    return $fullpath if $fullpath.e;
+    my @path = $*SPEC.splitdir: $fullpath.resolve.relative($!prefix).IO;
+    my $subdir = $!prefix;
+    for @path -> $dir {
+        die "Can only create dirs under \$.prefix, but '" ~ $path ~ "' is outside of it" if $dir eq '..';
+        $subdir .= add($dir);
+        next if $subdir.e;
+        $subdir.mkdir;
+    }
+    $subdir
+}
+
+method rmsubdir(IO:D() $entry) {
+    return unless $entry.e;
+    if $entry.d {
+        for $entry.dir -> $subentry {
+            next if $subentry.basename eq '..' | '.';
+            if $subentry.d {
+                self.rmsubdir($subentry);
+            }
+            else {
+                $subentry.unlink;
+            }
+        }
+        $entry.rmdir;
+    }
+    else {
+        $entry.unink;
+    }
+}
+
+method make-dist(Str:D :$dist = 'MVDist', :%fields, Bool :$standalone, Bool :$include = True) {
+    my %dist-fields =
+        api => "0.0.1",
+        ver => "0.0.1",
+        auth => 'test:raku',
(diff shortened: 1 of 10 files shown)


vrurg:
Additional testing is needed for system collapsing and `{ "any": [ ... ] }` entries.

packages/S11-modules/multi-version/lib/MultiVer.rakumod
```diff
@@ -0,0 +1,142 @@
...
+    run $*EXECUTABLE.absolute, |@compiler-args, $script, :in, :out, :err
+}
+
+method run-standalone(Str:D $code is copy, :@compiler-args, Bool :$use-lib) {
+    fail "Standalone dist is not installed yet" unless $!standalone;
+    if $use-lib {
+        $code = self.use-libs ~ "\n#line 1 test-code.raku\n" ~ $code;
+    }
+    $!standalone-script.spurt: $code;
+    my $p = self.run: $!standalone-script, :compiler-args[
```
ugexe, line 123: this isn't closing any of the handles it opens

packages/S11-modules/multi-version/lib/MultiVer.rakumod
```diff
@@ -0,0 +1,142 @@
...
+            }
+        }
+        $entry.rmdir;
+    }
+    else {
+        $entry.unink;
+    }
+}
+
+method make-dist(Str:D :$dist = 'MVDist', :%fields, Bool :$standalone, Bool :$include = True) {
```
ugexe, line 52: fwiw there is a similar utility function / pattern used elsewhere:
https://github.com/Raku/roast/blob/4ec7d433a47ee7344316959db9a0a6cdf90a918d/S11-repository/cur-candidates.t#L151-L159

packages/S11-modules/multi-version/lib/MultiVer.rakumod
```diff
@@ -0,0 +1,142 @@
...
+        auth => 'test:raku',
+        description => 'Test distribution',
+        |%fields;
+
+    my $dest-dir = $.prefix.add($dist ~ "-" ~ %dist-fields<ver>).IO;
+
+    # Remove the previous installation to guarantee no garbage is left behind.
+    self.rmsubdir($dest-dir);
+
+    for $?DISTRIBUTION.meta<resources>.grep({ .starts-with($dist ~ '/') && .ends-with('.in') }) -> IO() $tmpl {
```
ugexe, line 65: The exposed public interface is actually `$*DISTRIBUTION`.I have no idea why I used `$?DISTRIBUTION` elsewhere... I think originally it *was* `$?DISTRIBUTION` but then later I changed it to `$*DISTRIBUTION` (which still requires setting up a `$*DISTRIBUTION`)

packages/S11-modules/multi-version/lib/MultiVer.rakumod
ugexe: Actually I see this uses templates instead of generating everything

vrurg: I was considering it, but the particular needs of this test made it easier to implement own version.

packages/S11-modules/multi-version/lib/MultiVer.rakumod
vrurg: Documentation mentions `$?DISTRIBUTION`, the compiler installs it too. And it's the right name for it. With `$*DISTRIBUTION` there is a chance to pick up a wrong one from the call stack. Yet, the lexical nature of `$?DISTRIBUTION` makes it more error-proof.

Contrary, `$*DISTRIBUTION` is good for the compilation.

packages/S11-modules/multi-version/lib/MultiVer.rakumod
vrurg: Fixed with 0ef2346d4312686b00ae66ec563b41d3f18131dc. Thanks!

packages/S11-modules/multi-version/lib/MultiVer.rakumod
vrurg: > Actually I see this uses templates instead of generating everything

Somehow github didn't show this comment to me earlier. 

Yes, I use templates. And, BTW, it reminds me that I forgot about doing a control test for module version.

PR #828, opened 2023-01-27
Test for parameterised hashes as entries to store

2colours:
Enforces https://github.com/rakudo/rakudo/issues/5165 should not happen.


lizmat:
Thank you!

PR #830, opened 2023-02-19
Test implementing non-multi method stubbed in role

usev6:
... as multi method in class. Compare
https://github.com/Raku/old-issue-tracker/issues/2901

### DIFF
--- a/spectest.data
+++ b/spectest.data
@@ -742,6 +742,7 @@ S12-methods/typed-attributes.t
 S12-methods/what.t
 S12-subset/multi-dispatch.t
 S12-subset/subtypes.t
+S12-subset/type-subset.t
 S13-overloading/metaoperators.t
 S13-overloading/operators.t
 S13-overloading/typecasting-long.t


vrurg:
Considering that the error in issue is a compile-time one, wouldn't it be better to test using `eval-lives-ok`?

usev6:
The main idea of using ```eval-lives-ok``` would be, that there would be exactly one failing test instead of the whole test file blowing up when this would regress or when the tests are run with an old version? Or is there another advantage?

But anyway, your suggestion makes sense to me. I'll open a new pull request. (Also, I've messed this one up, but committing to my master branch.)

usev6:
New attempt at https://github.com/Raku/roast/pull/831

vrurg:
> The main idea of using `eval-lives-ok` would be, that there would be exactly one failing test instead of the whole test file blowing up when this would regress or when the tests are run with an old version? Or is there another advantage?

No, that's the only one.

PR #831, opened 2023-02-20
Test implementing non-multi method stubbed in role

usev6:
... as multi method in class. Compare
https://github.com/Raku/old-issue-tracker/issues/2901


vrurg (approved):

vrurg:
Thank you!

PR #837, opened 2023-07-30
Remove all :P5 related regex tests that were testing :P5

lizmat:
As suggested in:
  https://github.com/Raku/problem-solving/issues/378


zaucker:
Probably just a stupid comment by a reading only member of the group:

Wouldn't it make sense to first remove whatever implemented :P5 making sure all those tests fail and then remove the tests?

lizmat:
Not a stupid comment  :-)

The policy in Raku has always been that the spectest should be clean for any commit in Raku.  Now, of course that hasn't always been the case, but it was the goal.

Making first all tests fail, would be against this policy.

PR #838, opened 2023-09-09
Teach test_summary about #rakuast fudging

usev6:
This is similar to what was done in
https://github.com/rakudo/rakudo/commit/fa9cd8ef55.


lizmat:
Thanks!

PR #839, opened 2023-09-10
Update list of dirs for test statistics

usev6:
This reflects some recent changes of the directory structure for tests:

* Tests for Rosettacode have been removed in 2023-02, see https://github.com/Raku/roast/commit/504f0265ee
* Specific tests for 6.c and 6.d have been moved to dedicated dirs in https://github.com/Raku/roast/commit/720a5dc1f8 and follow-up commits

The change to the list of dirs doesn't really change the functionality of the script. It only avoids some warnings/errors regarding unknown or non-existing dirs. (One could argue that either the variable name or the logic of grouping tests per Synopsis should be updated, because the directory structure doesn't match the Synopsis names anymore. But maybe it's not worth the trouble.)

### DIFF
--- a/test_summary
+++ b/test_summary
@@ -68,7 +68,7 @@ for my $tfile (@tfiles) {
 
 # Prepare arrays and hashes to gather and accumulate test statistics
 my @col = qw(pass fail todo skip plan spec);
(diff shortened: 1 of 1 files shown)


usev6:
I went ahead and merged this. As mentioned, I don't think that this will have a negative effect. (There's always the "revert" button.)

PR #841, opened 2023-10-08
Fix tests for arrays that aren't shaped

usev6:
It looks like these tests have been added on the wrong assumption that adding a dot between the array name and the shape declaration should fail. At the time when the tests were added the synopsis (S09-data.pod) stated that those forms are errors. Later on it was clarified that they aren't supposed to die, but just can't be used to declare shaped arrays: https://github.com/Raku/old-design-docs/commit/e8527c2291.


usev6:
I'm pretty sure the tests were wrong, so I'll go ahead and merge this without a review.

PR #847, opened 2023-10-21
README: Added Our Contributors 

Dev7g:


README.md
```diff
@@ -159,3 +159,9 @@ too long a time by default so as to make the roast run faster.  Defaults to `1`.
 ## Contributing
 
 See [CONTRIBUTING.md](CONTRIBUTING.md)
+
+# Our Amazing Contributors ⭐
+
+<a href="https://github.com/vikhyatsingh123/Naruto-Shippuden/graphs/contributors">
```
usev6, line 165: The links look wrong. Probably a copy-pasto?

But even with that adjusted, I'm not convinced that we should add this block. I'm a bit skeptical about the dependency on contrib.rock and also the heading feels a bit obtrusive to me. But that's just my personal opinion.

Maybe adding a ```CREDITS``` file (as we do for [Rakudo](https://github.com/rakudo/rakudo/blob/main/CREDITS) and [NQP](https://github.com/Raku/nqp/blob/main/CREDITS)) and mentioning that here would be an alternative?

JJ: The user that's created the PR is not even part of the contributors of that other file linked, and none have anything to do whatsoever with this repo. This looks like a spam prove, and definitely not anything to do with this specific repository. The user has no track record on the Raku community, and looking at their contribution graph, it seems like they're looking to have some hacktoberfest fu accrued, after creating the account 6 days ago.
If it walks like a duck, quacks like a duck, dives like a duck, it *most probably* is a duck.
Might or might not add a CREDITS here, but that has absolutely nothing to do with this, which is probably spam, and that is why I blocked it.

JJ: The [site itself it's pointing to](https://github.com/vikhyatsingh123/Naruto-Shippuden) looks simply as an excuse to create easy PRs for Hacktoberfest.

2colours: I really enjoyed Naruto Shippuden though 🤣

usev6: BTW, I think the PR really wanted to look similar to this one https://github.com/FMC-Weekend/fmc24-client/pull/225 -- but pointing to https://github.com/Raku/roast/graphs/contributors of course.

PR #848, opened 2023-11-08
Improvements to `is default` tests for hashes

vrurg:
- Test all variants of hash parameterization of key and value.
- Fix tests for `is default` type check failures on attributes.


vrurg:
In support of rakudo/rakudo#5465

PR #850, opened 2023-12-13
Add test for instantiation of a generic class

vrurg:
In support of rakudo/rakudo#5493


JJ (approved):
Lines 20 to 31 could be DRYed, but LGTM.

PR #851, opened 2024-01-21
Test return prioritization

patrickbkr:
Discussed in https://github.com/Raku/problem-solving/issues/417

Depends on https://github.com/Raku/problem-solving/issues/417


vrurg (approved):

S06-advanced/return-prioritization.t
raiph, line 1: I'm confused about this test:
```
is s2(), 1, "LEAVE with return does not interrupt implicit return in different lexical scope";
```
First, the "implicit return" confuses me. What implicit return?

Second, the code appears to be more or less a duplicate of the earlier `s2(), 1, "LEAVE in other lexical scope can't interrupt return";`.  (Putting aside the confusing "implicit return". In both tests an *explicit* return doesn't get interrupted.)

Third, putting the "implicit" aside, I'd say the message of that *earlier* test would be better if it was `"LEAVE with return in other lexical scope can't interrupt return"`.

patrickbkr: Thanks for the review! I agree in all points. I'll fix this in the evening.

patrickbkr: Fixed.

PR #870, opened 2025-02-15
Fix tests for nested loops and labeled last

usev6:
As pointed out by @niner++ the old version assumed that the loops were not lazy. But as mentioned in the comment the old design docs suggested that the used syntax (having the loop inside parens) would make the loop lazy -- at least if the loop is the final statement of a statement list.

This behavior is currently neither fully specced nor implemented by Rakudo. The tests weren't meant to specify the non-lazy evaluation. They only purpose was to make sure that the code doesn't crash and that the label is taken into account.

### DIFF
--- a/S04-statements/label.t
+++ b/S04-statements/label.t
@@ -89,41 +89,119 @@ throws-like { EVAL q[label1: say "OH HAI"; label1: say "OH NOES"] }, X::Redeclar
 # NB(bartolin) As of 2020-04-13 Rakudo has different classes for different loop
 #              types. The tests resemble Rakudo's current internal structure,
 #              but the code should work for all implementations.
+#              Please note that each type of loop comes in two versions: not
+#              within parens and within parens. Accordings to the old design
+#              documents the second version should be lazy:
+#              https://github.com/Raku/old-design-docs/blob/a4c36c683d/S04-control.pod#loops-at-the-statementlist-level-vs-the-statement-level
+#              The tests explicitly don't test this aspect. They are about the
+#              crash described in the mentioned bug report.
 {
+    my $ignore;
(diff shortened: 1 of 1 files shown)


usev6:
There is a problem when running with `RAKUDO_RAKUAST=1` with one of the tests:
```
$ RAKULIB=lib RAKUDO_RAKUAST=1 ./rakudo-m t/spec/S04-statements/label.t
[...]
not ok 9 - no crash: nested loops and labeled last (1)
# Failed test 'no crash: nested loops and labeled last (1)'
# at t/spec/S04-statements/label.t line 118
# expected: $[1, 3, 5, 7]
#      got: $[1, 5]
[...]
WARNING: unhandled Failure detected in DESTROY. If you meant to ignore it, you can mark it as handled by calling .Bool, .so, .not, or .defined methods. The Failure was:
Index out of range. Is: 2, should be in 0..0
  in block <unit> at t/spec/S04-statements/label.t line 117

WARNING: unhandled Failure detected in DESTROY. If you meant to ignore it, you can mark it as handled by calling .Bool, .so, .not, or .defined methods. The Failure was:
Index out of range. Is: 2, should be in 0..0
  in block <unit> at t/spec/S04-statements/label.t line 115

```

The Failure seems to happen, because no Seq is returned with `RAKUDO_RAKUAST=1`:
```
$ ./rakudo-m -e 'dd sub () { (L: while True { last L }) }()'
(Label.new(name => "L", file => "-e", line => 1),).Seq
$ RAKUDO_RAKUAST=1 ./rakudo-m -e 'dd sub () { (L: while True { last L }) }()'
Label.new(name => "L", file => "-e", line => 0)
```

PR #871, opened 2025-03-14
Simplify 6.c/s14 roles/mixin 6c.t

MasterDuke17:


6.c/S14-roles/mixin-6c.t
```diff
@@ -125,10 +125,11 @@ is $y.test,     42,         'method from other role was OK too';
 # https://github.com/Raku/old-issue-tracker/issues/2943
 {
     my $rt115390 = 0;
+    my $sum = 0;
     for 1..1000 -> $i {
         $rt115390 += $i.raku;
         my $error = (my $val = (^10).pick(3).min but !$rt115390);
-        1
+        $sum++;
```
niner, line 132: You can just replace the 1 with a Nil

MasterDuke17: Doh.

PR #900, opened 2026-08-19
Reach the run time type check with a value the compiler cannot judge

ugexe:
Rakudo is learning to report a store of a value a variable's type can never accept while it compiles, instead of leaving it to the run time check. A test written as `dies-ok { my Int $x = "str" }` names a literal the compiler can judge, so it stops reaching the check it was written for.

Each of these holds the value in a variable instead. The store still happens, the type still rejects it, and the check under test is the one that runs. The two subs returning a typed positional get values written as the type their array declares, so what fails is the return constraint they are testing rather than the declaration above it.

### DIFF
--- a/6.c/S04-declarations/my-6c.t
+++ b/6.c/S04-declarations/my-6c.t
@@ -235,23 +235,27 @@ throws-like 'my $z = $z', X::Syntax::Variable::Initializer, name => '$z';
     # If there is a regression this may die not just fail to make ints
     eval-lives-ok 'my (int $a);','native in declarator sig';
     eval-lives-ok 'my (int $a, int $b);','natives in declarator sig';
-    dies-ok { my (int $a, num $b); $a = 'omg'; }, 'Native types in declarator sig 1/2 constrains';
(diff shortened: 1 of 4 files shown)


6.c/S04-declarations/my-6c.t
```diff
@@ -235,23 +235,27 @@ throws-like 'my $z = $z', X::Syntax::Variable::Initializer, name => '$z';
     # If there is a regression this may die not just fail to make ints
     eval-lives-ok 'my (int $a);','native in declarator sig';
     eval-lives-ok 'my (int $a, int $b);','natives in declarator sig';
-    dies-ok { my (int $a, num $b); $a = 'omg'; }, 'Native types in declarator sig 1/2 constrains';
-    dies-ok { my (int $a, num $b); $b = 'omg'; }, 'Native types in declarator sig 2/2 constrains';
+    # The value assigned is held in a variable so that the declared type is what
+    # rejects it, rather than the compiler rejecting the literal out of hand.
+    my $omg = 'omg';
```
lizmat: Feels to me that all tests using `$omg` can also be done with`$str`, so why 2 different variables with a string?

